PR #35, opened 2016-04-07
point users to consolidated install docs

bsweger:
@kevinli-work or @mab6bf -- small change to point folks to the consolidated install docs


kevinli-work:
Should the main README document the various install options? The `data-act-core` repo docs don't seem to have all the info on the various build options the old frontend README did.

bsweger:
@kevinli-work thanks for reviewing! I'm definitely open to making the directions as complete and clear as possible. I left out the info about various build options because the consolidated install docs are targeted to local developers. Is it correct to assume that audience would be working from the _dev config file?

As I followed along to set up my own local site, I found the various options in the current directions to be a bit confusing. My personal preference is to only see the info necessary to get up and running for development. That said, I don't want to omit any meaningful nuance. 

For the other repos, we've left some of the detailed info in the REAMDEs. Does it make sense to follow that model here (i.e., leave info about the various config options in the web-app repo)?

PR #61, opened 2016-04-15
Help docs

kaitlin:
Updated the add data page with dummy files and a bit more explanation of what they'll need.

cc @kevinli-work


kevinli-work:
Merging into wrong branch

PR #97, opened 2016-05-03
Hotfix: Saner gulp output

kevinli-work:
* Gulp will now always output to `public/` and use the same folder structure, even in local dev environments


kevinli-work:
Wrong branch

PR #105, opened 2016-05-09
updated Help Content 

kaitlin:
per @lisaam 's edits in PUB-124.

BEFORE MERGING @kevinli-work can you update the RSS link on the help page with the newest version pinned in slack and push to this branch?

### DIFF
--- a/src/js/components/help/helpContent.jsx
+++ b/src/js/components/help/helpContent.jsx
@@ -55,9 +55,9 @@ export default class HelpContent extends React.Component {
     render() {
         return (
             <div className="usa-da-help-content">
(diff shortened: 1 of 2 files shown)


lisaam:
Originally, I was looking to have the mailto link at the bottom of the page updated. Today I received more edits, including update the RSS link to https://community.max.gov/x/9IDQPQ (per Mindy Levit).

kaitlin:
I think it makes sense to leave the RSS link to the broker as is because 

a) we don't want the MAX one to update without us knowing and it becomes out of sync with the broker
b) they will have to log in to MAX to view it, after logging in to the broker

If there are more text edits please send them along and I'll incorporate. 

The mail to links were updated as they appear in the doc you sent. Are they incorrect here?

PR #140, opened 2016-06-01
Updated help page content

kaitlin:
Includes updated nav and help page content for sprint 13


kevinli-work:
Made slight HTML tag changes to prevent React from returning errors/warnings:
* `<p>` tags cannot wrap around `<ul>` tags
* `<td>` tags must be nested inside a `<tbody>` tag

PR #158, opened 2016-06-15
Help Page

kevinli-work:
* Developed an inline-HTML parser to handle custom Markdown syntax indicating raw HTML to be inserted into page
* Restyled tables so they don't overflow off the redesigned help page
* Includes latest help text from @lisaam


nmonga91:
Looks good, merging

PR #195, opened 2016-07-12
Update links

kevinli-work:


bsweger:
@kevinli-work This looks good--thanks for the change!

PR #207, opened 2016-07-26
update for sprint 17

lisaam:
Includes reworked rules table - please QA carefully. not sure I used class="break-word" correctly


kevinli-work:
Included in #209

PR #228, opened 2016-08-04
Help Links

kevinli-work:
* Removed all API calls to `/get_rss`
* API integration with `/get_protected_files`
* Removed Resources section from `changelog.md`, implemented it with signed URLs in JS
* Added link to new Resources section in validation table
* Removed Submission Guide/Add Data Webpack chunk; these are now bundled into the initial `app.js` file to prevent the Submission Guide from appearing momentarily after the user has chosen to hide it

Merge approvals:
- [x] Code reviewed
- [x] @nmonga91 sign off

### DIFF
--- a/src/js/components/help/helpContent.jsx
+++ b/src/js/components/help/helpContent.jsx
@@ -5,7 +5,7 @@
 
 import React from 'react';
 import $ from 'jquery';
-import { generateRSSUrl, generateProtectedUrls } from '../../helpers/util.js';
+import { generateRSSUrl, generateProtectedUrls, rssFileKey } from '../../helpers/util.js';
 
 
 let gifSrc = 'graphics/reportabug.gif';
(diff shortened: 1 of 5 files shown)


src/js/components/help/helpContent.jsx
```diff
@@ -16,31 +16,40 @@ export default class HelpContent extends React.Component {
...
-        this.rssPromise.promise
-            .then((url) => {
+
+        // also load the remaining URLs
+        this.urlPromise = generateProtectedUrls();
+        this.urlPromise.promise
+            .then((urls) => {
                 this.setState({
-                    rssUrl: url
+                    rssUrl: urls['RSS_v1.0.xlsx'],
```
nmonga91: So the endpoint is returning the keys in the response as the filename of whatever is in the S3 bucket for the static files. If it doesn't match to what's here, it'll think it wasn't part of the response. I think we need a better way to handle it.

kevinli-work:
Fair, but we _have to_ know what the file name is, either in the frontend or the backend, at some point to know what files to retrieve. Are you proposing that the backend track the file keys?

PR #250, opened 2016-09-02
Hotfix: JS syntax error

kevinli-work:
Added shrinkwrap


cmc333333:
The code fix resolves the problem for me locally, and `shrinkwrap`'s generally a good idea so :+1:

PR #252, opened 2016-09-02
Adds a simple travis test script

cmc333333:
It `buildLocal`s against three versions of node


kevinli-work:
@cmc333333 Is this something you want merged into dev or was this a test?

cmc333333:
@kevinli-work it can definitely be merged

PR #273, opened 2016-09-14
Promote date change to staging

kevinli-work:


klrBAH:
Tests expected to fail until travis config pushed past dev.

PR #326, opened 2016-10-17
changes for Oct 19 (to dev-stable)

lisaam:
@kaitlin to review content


kaitlin:
A-OK from a content perspective.

kevinli-work:
It seems that all the section links in the Release Notes Archive page still point to `/#/help?section=XXX`, resulting in broken links.

Besides updating the links to point to `/#/history`, does the history page contain the logic to jump to the correct sections? Is this relevant?

lisaam:
I don't think the history page needs to jump to each item.  Too many items

Future item: I did **just** have an idea that history page jumps to each release date section (like September 30) might be use full.

PR #338, opened 2016-10-21
MAX Login (Frontend)

kevinli-work:
* Integrates MAX SSO login functionality on the frontend.

- [x] Code review
- [x] @nmonga91 review
- [x]  Coordinated deployment with fedspendingtransparency/data-act-broker-backend#370


src/js/components/login/AuthPage.jsx
```diff
@@ -0,0 +1,33 @@
+/**
+* LoginPage.jsx
+* Created by Kyle Fox 12/4/15
```
mab6bf: May want to update these

src/js/components/login/LoginMax.jsx
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    detectRedirection() {
+        // check if the URL has a redirect param, save it in the state
+        if (this.props.location.query.hasOwnProperty('redirect')) {
+            this.setState({
+                redirect: this.props.location.query.redirect
+            }, () => {
+                // save the redirect destination as a cookie, expire after 5 min
+                Cookies.set('brokerRedirect', this.state.redirect, {expires: (5/(24*60))});
```
mab6bf: Perhaps we should add a comment that the expiration time is measured in days

src/js/components/login/LoginMax.jsx
```diff
@@ -0,0 +1,61 @@
...
+            });
+        }
+    }
+
+	render() {
+		return (
+                <div className="row">
+                    <div className="col-xs-12">
+                    	<p className="instructions">Sign in or register for the DATA Act Broker using your MAX ID.</p>
+                        <a href={kGlobalConstants.CAS_ROOT + '/cas/login?service=' + encodeURIComponent(kGlobalConstants.AUTH_CALLBACK)} className="usa-da-button btn-primary btn-lg btn-full">Sign In Using MAX</a>
```
mab6bf: Do we need to add any error handling in the case that CAS_ROOT or AUTH_CALLBACK aren't defined for local installs?

src/js/containers/login/AuthContainer.jsx
```diff
@@ -0,0 +1,107 @@
...
+
+	componentDidMount() {
+		this.processTicket();
+	}
+
+	processTicket() {
+
+		// extract the ticket string from the URL
+		const url = window.location.href;
+		// MAX may insert the ticket in the middle of the URL instead of at the end like a civilised person
```
mab6bf: Amusing, but let's clean this one up

src/js/components/login/LoginMax.jsx
kevinli-work: Local installs use the old login approach https://github.com/fedspendingtransparency/data-act-broker-web-app/blob/feature-maxsso/src/js/containers/login/LoginContainer.jsx#L53

kevinli-work:
Comments about the MAX return URL format is necessary because using a regex to extract the token is an antipattern; we _should_ be able to do it with React Router which has URL query param parsing built-in (and one would assume with better edge case handling). MAX presumably assumes that hashes in URLs are traditional in-page anchors rather than SPA virtual routing, so it inserts the token as a query param before the hash, which makes the params inaccessible to React Router.

nmonga91:
One question: when the session is killed and we have to login with MAX to continue what we're doing, we don't come back to the page we were originally on. I know that the redirect URL we're sending to MAX always redirects to our main splash page. The intention was to handle that dynamic redirect from MAX later on, correct?

kevinli-work:
^ this is now dealt with

PR #349, opened 2016-11-14
Removing permissions handling

nmonga91:
Removing permissions handling since it is no longer managed via the broker but instead through MAX.

- [x] Reviewed by front-end
- [x] Merge coordinated with https://github.com/fedspendingtransparency/data-act-broker-backend/pull/406


src/js/helpers/loginHelper.js
```diff
@@ -32,11 +32,9 @@ export const fetchActiveUser = () => {
 
 	                // check to see if the user is an admin
 	                let isAdmin = false;
-	                res.body.permissions.forEach(function(perm) {
-	                    if (perm == 1) {
-	                        isAdmin = true;
-	                    }
-	                });
+	                if (res.body.permission == 4) {
```
kevinli-work: Is the API response going to change to a single `permission` value instead of returning an array?

PR #358, opened 2016-11-18
Stac sch patch 1

StacSch:
Replace 10/31/2016 versions with 11/23/16 versions of changelog.md and history.md.

@ kaitlin


cmc333333:
@StacSch this includes some commits we don't want just yet. I'll PR just the text updates shortly.

PR #370, opened 2016-11-21
Reverting resource page content change to commit id 881b65b

nmonga91:
Reverting content changes per the request of @lisaam


cmc333333:
Alrighty. Low risk and I don't want this to hold up deploying, so I'm going to merge.

PR #374, opened 2016-11-22
Db 1382 review comments

alburde1:
This pull request allows the user to add comments to each file and save the changes on the review screen.


cmc333333:
This is working well for me, but I don't feel like a good enough judge of code quality here to merge.

src/js/components/reviewData/ReviewDataComments.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+        this.state = {
+            currentFile: "A",
+            fileComments: props.comments,
+            saveState: ""
+        };
+    }
+
+    changeFile(newFile) {
+        var fileComment = document.getElementById("submission-review-comments").value.trim();
```
cmc333333: Apologies if this is a bad suggestion, but it seems like we shouldn't be using `document` in these functions. Is this where `refs` should be used?

src/js/containers/review/ReviewDataContainer.jsx
```diff
@@ -56,6 +56,10 @@ class ReviewDataContainer extends React.Component {
             })
             .then((name) => {
                 submission.agency_name = name;
+                return ReviewHelper.fetchSubmissionComments(this.props.params.submissionID);
+            })
+            .then((comments) => {
+                submission.file_comments = comments;
```
cmc333333: Does it make sense to async all these things in parallel? The main data, obligations, and narratives don't depend on each other.

src/js/components/reviewData/ReviewDataComments.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+import * as ReviewHelper from '../../helpers/reviewHelper.js';
+
+export default class ReviewDataComments extends React.Component {
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            currentFile: "A",
+            fileComments: props.comments,
```
kevinli-work: Let's not mix our props and states. Additionally, in an async situation, the component may mount before the server responds with the server prop values.

src/js/components/reviewData/ReviewDataComments.jsx
kevinli-work: @cmc333333 is correct; we rarely directly touch the DOM in React because it is not guaranteed to exist (or exist in the way you think it will) when the code executes. One alternative would be to convert `ReviewDataCommentsTextfield` to be controlled by `ReviewDataComments` (ie, have the `textarea`'s value stored in `ReviewDataComments`'s state instead of in `ReviewDataCommentsTextfield`) so you can grab the text value out of the state instead of reading from DOM.

You _could_ use refs per CM's suggestion, but the fact that you're cloning props into state in child components suggests the current approach is an anti-pattern and that the master state should be in a higher component. Remember, you can flow changes upward to parent components via functions that are passed down through props, but prop data should represent data that isn't changed by the component itself.

src/js/components/reviewData/ReviewDataComments.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+        this.setState({
+            fileComments: tempComments,
+            currentFile: newFile,
+        });
+    }
+
+    saveComments() {
+        this.setState({saveState: "Saving"});
+        var fileComment = document.getElementById("submission-review-comments").value.trim();
```
kevinli-work: Same issues as above; also, this is duplicated code. Could you consolidate the logic of saving the comments to state in a single, reusable function?

src/js/containers/review/ReviewDataContainer.jsx
kevinli-work: I'm fine with keeping these as chained requests because we need to fetch the agency name after we get the CGAC code back from the main data anyway. Parallel promises (for the obligations and narratives) uses a sufficiently different syntax and approach that it may not be worth mixing the two.

PR #400, opened 2016-12-14
Dev stable -> dev

nmonga91:


cmc333333:
Odd about the push failing, but the PR's the one that matters.

PR #403, opened 2016-12-15
Db 1528 remove unused routes

alburde1:
This PR is PROBABLY easier to look at one commit at a time, that way it's at least broken up by type.

This PR removes the routes/styles/components for the forgot password, registration, and admin pages. These are no longer needed with MAX login. It also cleans up a couple extra components we weren't using anymore and a couple imports in files that weren't removed.


alburde1:
[run tests]

mab6bf:
[run tests]

PR #411, opened 2016-12-19
Add eslintrc rules consistent w/ codebase + Travis

cmc333333:
Rather than go without a linter (our current strategy), let's draw a line in
the sand and prevent future issues by adding a linter that matches our current
code base. We're breaking some pretty important rules (e.g. no-undef), so we
should revisit these to fix things ASAP.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -10,7 +10,7 @@
         "comma-spacing": 0,
         "dot-notation": 0,
         "eol-last": 0,
-        "eqeqeq": 0,
+        "eqeqeq": 0,    // ASAP
         "guard-for-in": 0,
         "indent": 0,
         "key-spacing": 0,
@@ -21,16 +21,15 @@
         "no-multi-spaces": 0,
         "no-param-reassign": 0,
         "no-trailing-spaces": 0,
-        "no-undef": 0,
+        "no-undef": 0,    // ASAP
         "no-unused-vars": 0,
-        "no-use-before-define": 0,
+        "no-use-before-define": 0,  // ASAP
         "object-curly-spacing": 0,
         "object-shorthand": 0,
         "padded-blocks": 0,
         "prefer-const": 0,
         "quote-props": 0,
         "quotes": 0,
-        "radix": 0,
         "semi": 0,
         "space-infix-ops": 0,
     }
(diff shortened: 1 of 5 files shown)


.eslintrc
```diff
@@ -2,7 +2,34 @@
     "extends": "airbnb",
     "rules": {
         // Override any settings from the airbnb configuration
-        "indent": [2, 4],
-        "comma-dangle": 0
+        // "indent": [2, 4],
```
nmonga91: is this override being removed because the default is sufficient?

.eslintrc
```diff
@@ -2,7 +2,34 @@
...
+        "comma-dangle": 0,
+        // To fix
+        "brace-style": 0,
+        "camelcase": 0,
```
nmonga91: Does `0` indicate "don't override and use the default" ?

src/js/helpers/reviewHelper.js
```diff
@@ -172,7 +172,7 @@ const getFileStates = (status) => {
 		else {
 			let count = 0;
 			item.error_data.forEach((error) => {
-				count += parseInt(error.occurrences);
+				count += parseInt(error.occurrences, 10);
```
nmonga91: The `10` here is to ensure the result is in base 10. But that should be the default (provided that the number in question doesn't start with a 0), no? 
Is it possible for `error.occurrences` to start with a 0?

.eslintrc
cmc333333: Nah, I commented it out (and turned it off later) because our current code doesn't match this style.

.eslintrc
cmc333333: `0` is the same as `off`, it means "ignore this rule". So we're turning off rules that were turned on in the airbnb style guide.

src/js/helpers/reviewHelper.js
cmc333333: Good question! Adding the `base` parameter to `parseInt` is just a sanity check. We almost never want to parse an integer in a non-10 base (and when we do, we should be explicit that that's the intent), so one of the [linting rules](http://eslint.org/docs/rules/radix) checks for this. I think that's a reasonable standard, but would be happy to revert if there's contention.

nmonga91: I think we should keep it if the linting rules are checking for it. Being explicit never hurt so I'm definitely ok with leaving it in.

nmonga91:
Looks good, merging

PR #420, opened 2016-12-30
Include submission context in UI

cmc333333:
~~Depends on fedspendingtransparency/data-act-broker-backend#484~~. Adds the agency name and submission date to the floating submission context:

screen shot 2016-12-30 at 12 02 54 pm

### DIFF
--- a/src/js/components/addData/AddDataHeader.jsx
+++ b/src/js/components/addData/AddDataHeader.jsx
@@ -8,15 +8,15 @@ import moment from 'moment';
 
 import * as ReviewHelper from '../../helpers/reviewHelper.js';
 
-class LastUpdated extends React.Component {
-    constructor(props){
-        super(props);
-    }
-    
+class SubmissionContext extends React.Component {
     render() {
         return (
             <div className="last-updated">
-                Last Saved: {this.props.last_updated}
+                Last Saved: {this.props.formattedTime}
+                <br />
+                {this.props.agencyName}
+                <br />
+                {this.props.submissionTime}
             </div>
         );
     }
@@ -60,10 +60,14 @@ export default class AddDataHeader extends React.Component {
     }
 
     render() {
-        let lastUpdated = null;
-        if (this.state.ready && this.state.last_updated){
+        let submissionContext = null;
+        if (this.state.ready) {
             let formattedTime = moment.utc(this.state.last_updated).local().format('h:mm a');
-            lastUpdated = <LastUpdated last_updated={formattedTime} />
+            submissionContext = <SubmissionContext
+              formattedTime={formattedTime}
+              agencyName={this.state.agency_name}
+              submissionTime={this.state.reporting_period_start_date}
+            />
         }
 
         return (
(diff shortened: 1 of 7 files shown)


src/js/components/addData/AddDataHeader.jsx
```diff
@@ -8,15 +8,15 @@ import moment from 'moment';
...
+class SubmissionContext extends React.Component {
     render() {
         return (
             <div className="last-updated">
-                Last Saved: {this.props.last_updated}
+                Last Saved: {this.props.formattedTime}
+                <br />
+                {this.props.agencyName}
+                <br />
+                {this.props.submissionTime}
```
nmonga91: `submissionTime` corresponds to `Q3/2016` in the screenshot, correct? If so, should the variable be named something like `submissionFiscalTimePeriod` as to not confuse it with a timestamp?

src/js/components/addData/AddDataHeader.jsx
```diff
@@ -60,10 +60,14 @@ export default class AddDataHeader extends React.Component {
...
-        let lastUpdated = null;
-        if (this.state.ready && this.state.last_updated){
+        let submissionContext = null;
+        if (this.state.ready) {
             let formattedTime = moment.utc(this.state.last_updated).local().format('h:mm a');
-            lastUpdated = <LastUpdated last_updated={formattedTime} />
+            submissionContext = <SubmissionContext
+              formattedTime={formattedTime}
+              agencyName={this.state.agency_name}
+              submissionTime={this.state.reporting_period_start_date}
```
nmonga91: A user can also select a specific month instead of a quarter for a fiscal year, but we're using the `reporting_period_start_date` to convert it to Quarter/FY format right? Do we need to account for the selection of a specific month?

src/js/containers/review/ReviewDataContainer.jsx
```diff
@@ -8,7 +8,6 @@ import { bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
 
 import * as ReviewHelper from '../../helpers/reviewHelper.js';
-import * as AgencyHelper from '../../helpers/agencyHelper.js';
```
nmonga91: Removing this here removes the Agency Name info from the Review & Certify page correct?

src/js/components/addData/AddDataHeader.jsx
cmc333333: `submissionTime` will be `Q3/2016` for quarterly submissions as a date for month submissions, so I'm not sure of a great variable name. `timePeriodLabel`?

src/js/components/addData/AddDataHeader.jsx
cmc333333: Per the previous comment, there's no conversion on the front end.

src/js/containers/review/ReviewDataContainer.jsx
cmc333333: Not quite -- `AgencyHelper` did an AJAX call to grab a list of agency names and ids and then use the submission's `cgac_code` to look up an agency name. As the submission now includes the agency name directly, there's no need for the AJAX step.

src/js/components/addData/AddDataHeader.jsx
nmonga91: `timePeriodLabel` sounds good!

PR #422, opened 2017-01-04
Replace signErrorWarningReport w/ submissionReport

cmc333333:
This avoids creating the file name within JS for security and flexibility. It must be merged concurrently with fedspendingtransparency/data-act-broker-backend#489


nmonga91:
So in general, we're using `reportWarning` as a boolean. If true, grab the warning report, else default to the error report?

cmc333333:
Exactly. We could pass around a string to represent warning vs error, but
that's just another input to validate.

PR #433, opened 2017-01-27
Db 1555 recent activity limit

kimminahm:


src/js/components/landing/LandingContent.jsx
```diff
@@ -55,12 +55,41 @@ export default class LandingContent extends React.Component {
...
+        //Size to set off the limiting of the recent activity
+        let limit=3;
+
         let agencyName = 'Your Agency';
-        if (affiliations && affiliations.length > 0) {
+        if (affiliations && affiliations.length > 0 && this.state.expanded) {
+            agencyName = affiliations.map((a) => a.agency_name).join(', ');
+        }else if(affiliations && affiliations.length > limit && !this.state.expanded){
+            agencyName = affiliations.map((a) => a.agency_name)[0] + "...";
+        }else if(affiliations && affiliations.length < limit && affiliations.length > 0 && !this.state.expanded){
```
alburde1: Could the last "else if" and the first "if" be combined? They end up doing the same thing. Maybe put the second "if" as the first "if" and then make an else of just `affiliations && affiliations.length > 0`.

So it would look like this:
```
if(affiliations && affiliations.length > limit && !this.state.expanded){
	agencyName = affiliations.map((a) => a.agency_name)[0] + "...";
}
else if (affiliations && affiliations.length > 0) {
	agencyName = affiliations.map((a) => a.agency_name).join(', ');
}
```

alburde1: Also, with your set of checks, what happens if the state is not expanded but the limit is equal to the length? The shorter version covers that issue by not caring what the state is in the second if.

src/js/components/landing/LandingContent.jsx
```diff
@@ -55,12 +55,41 @@ export default class LandingContent extends React.Component {
...
+        this.setState({
+            expanded: !this.state.expanded
+        })
+    }
+
     render() {
         const affiliations = this.props.session.user.affiliations;
+
+        //Size to set off the limiting of the recent activity
+        let limit=3;
```
alburde1: Since limit isn't changing, you should use `const` here instead of `let`

PR #437, opened 2017-02-03
Db 1602 mk upgrade max permissions

kimminahm:
HelpOnly users will have access to a link that directs them to max to increase their permissions.

- Users without HelpOnly permissions will not see any changes on the help page
- Users with HelpOnly permissions will see a link in the sidebar below 'Getting More Help' which autoscrolls to a heading with a link to https://community.max.gov/x/fJwuRQ

### DIFF
--- a/src/js/components/help/helpContent.jsx
+++ b/src/js/components/help/helpContent.jsx
@@ -21,7 +21,7 @@ export default class HelpContent extends React.Component {
         this.state = {
             rssUrl: '',
             validationRulesUrl: '#',
-            domainValuesUrl: '#',
+            domainValuesUrl: '#'
         };
     }
 
--- a/src/js/components/help/helpContent.jsx
+++ b/src/js/components/help/helpContent.jsx
@@ -21,7 +21,7 @@ export default class HelpContent extends React.Component {
         this.state = {
             rssUrl: '',
             validationRulesUrl: '#',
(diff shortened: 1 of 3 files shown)


src/js/components/help/helpContent.jsx
```diff
@@ -62,6 +62,15 @@ export default class HelpContent extends React.Component {
     }
 
     render() {
+        let membership = '';;
```
alburde1: There are 2 semicolons here

alburde1: Also, when setting something to be empty (not show up), it's better to set it to null. If you set it to an empty string React inserts a <span></span> there anyway. If you set it to null it doesn't insert anything, which is cleaner and better overall (also potentially less problematic for accessibility issues).

src/js/components/help/helpSidebar.jsx
```diff
@@ -7,11 +7,25 @@ import React from 'react';
...
+    constructor(props) {
+        super(props);
+    }
+
     render() {
         const sectionList = this.props.sections.map((section, index) => {
             return <HelpSidebarItem key={index} sectionName={section.name} sectionId={section.link} />
         });
 
+        let membership = '';
```
alburde1: Same as above, should be `null` rather than `''`

PR #440, opened 2017-02-06
Db 1624 mk download uploaded files

kimminahm:
Added download link. Tested by uploading file and downloading the uploaded file.

### DIFF
--- a/src/js/components/validateData/validateValues/ValidateValuesFileComponent.jsx
+++ b/src/js/components/validateData/validateValues/ValidateValuesFileComponent.jsx
@@ -28,7 +28,6 @@ const propTypes = {
 export default class ValidateDataFileComponent extends React.Component {
     constructor(props) {
         super(props);
-        console.log(props);
         this.state = {
             showWarning: false,
             showError: false,
@@ -39,7 +38,7 @@ export default class ValidateDataFileComponent extends React.Component {
 
     componentDidMount() {
         this.determineErrors(this.props.item);
-        this.getFileUrl(this.props.item);
+        // this.getFileUrl(this.props.item);
     }
 
     componentWillReceiveProps(nextProps) {
@@ -86,18 +85,19 @@ export default class ValidateDataFileComponent extends React.Component {
         return stagedFile;
     }
 
-    getFileUrl(item){
+    downloadFile(item){
+        console.log(item);
         if(item.file_type=='appropriations'){
             GenerateFilesHelper.fetchFile('A', this.props.submission.id).then((result)=>{
-                this.setState({url: result.url});
+                window.location = result.url;
             })
         }else if(item.file_type=='program_activity'){
             GenerateFilesHelper.fetchFile('B', this.props.submission.id).then((result)=>{
-                this.setState({url: result.url});
+                window.location = result.url;
             })
         }else if(item.file_type=='award_financial'){
             GenerateFilesHelper.fetchFile('C', this.props.submission.id).then((result)=>{
-                this.setState({url: result.url});
+                window.location = result.url;
             })
         }
     }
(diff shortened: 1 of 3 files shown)


src/js/components/validateData/validateValues/ValidateValuesFileComponent.jsx
```diff
@@ -27,7 +28,7 @@ const propTypes = {
 export default class ValidateDataFileComponent extends React.Component {
     constructor(props) {
         super(props);
-
+        console.log(props);
```
alburde1: Extraneous console log

src/js/components/validateData/validateValues/ValidateValuesFileComponent.jsx
```diff
@@ -84,6 +86,22 @@ export default class ValidateDataFileComponent extends React.Component {
...
+            GenerateFilesHelper.fetchFile('B', this.props.submission.id).then((result)=>{
+                this.setState({url: result.url});
+            })
+        }else if(item.file_type=='award_financial'){
+            GenerateFilesHelper.fetchFile('C', this.props.submission.id).then((result)=>{
+                this.setState({url: result.url});
+            })
+        }
+    }
+
```
alburde1: The problem with this method of setting the url is that the url can expire if the user sits on this page for too long and they have to refresh to be able to access it again. Take a look at `components/validateData/validateValues/ValidateValuesErrorReport.jsx` for a sample of how to get the URL on click.

I know we use this method in other places as well, but not all of them have been updated to generate the URL on click so some URLs can still expire

src/js/components/validateData/validateValues/ValidateValuesFileComponent.jsx
```diff
@@ -38,6 +38,7 @@ export default class ValidateDataFileComponent extends React.Component {
 
     componentDidMount() {
         this.determineErrors(this.props.item);
+        // this.getFileUrl(this.props.item);
```
alburde1: Don't leave the commented-out stuff here ;)

src/js/components/validateData/validateValues/ValidateValuesFileComponent.jsx
```diff
@@ -27,7 +28,6 @@ const propTypes = {
 export default class ValidateDataFileComponent extends React.Component {
     constructor(props) {
         super(props);
-
```
alburde1: Might as well leave this space in for readability (not a big deal but since I've already asked for a different change, might as well update this while you're in there)

src/js/components/validateData/validateValues/ValidateValuesFileComponent.jsx
```diff
@@ -84,6 +85,23 @@ export default class ValidateDataFileComponent extends React.Component {
         return stagedFile;
     }
 
+    downloadFile(item){
+        console.log(item);
```
alburde1: Console log

src/js/components/validateData/validateValues/ValidateValuesFileComponent.jsx
```diff
@@ -84,6 +85,23 @@ export default class ValidateDataFileComponent extends React.Component {
...
+        }else if(item.file_type=='program_activity'){
+            GenerateFilesHelper.fetchFile('B', this.props.submission.id).then((result)=>{
+                window.location = result.url;
+            })
+        }else if(item.file_type=='award_financial'){
+            GenerateFilesHelper.fetchFile('C', this.props.submission.id).then((result)=>{
+                window.location = result.url;
+            })
+        }
+    }
```
alburde1: You might want to include a catch for errors. Also, some sign that it's working (like in the sample I pointed you to with the `signInProgress` example) might be nice. If it's taking a while people might click multiple times. We only want to sign it if we aren't already in the process of doing so. Specifically look at `clickedReport` and `signReport` for what I'm talking about.

PR #444, opened 2017-02-09
Db 1628 mk technical notes

kimminahm:
Must have a help/technical.md as well as help/technicalHistory.md to provide content for the page


src/help/technicalHistory.md
```diff
@@ -0,0 +1,48 @@
+#### January 18, 2017
```
alburde1: If we already have a history.md file, why does this file have the same contents? Do we now have 3 history files? 2 `history.md` files, one in `src/help` and one in `public/help` and then this `technicalHistory.md`?

src/help/technical.md
```diff
@@ -0,0 +1,140 @@
+#### December 7, 2016
```
alburde1: Why does this end before the history one if this is current updates? Shouldn't this be history?

src/js/helpers/helpHelper.js
```diff
@@ -137,6 +156,35 @@ export const loadHelp = () => {
...
+	};
+
+	loadTechnicalNotes()
+		.then((data) => {
+			output.html = data.html;
+			output.sections = data.sections;
+			return loadHistory();
+		})
+		.then((data) => {
+			output.history = data;
```
alburde1: Is this actually accessed anywhere? If not, do we even need this section? It seems like it wouldn't be since you only load history on the `/history` page, which is a new call to the md files and just loads `history.md`. `technicalHistory.md` isn't referenced anywhere at all.

src/help/technicalHistory.md
kimminahm: Dummy content to test that it is pulling files correctly. These will be modified manually be Kaitlin

src/help/technical.md
kimminahm: Dummy content for testing purposes

src/help/technicalHistory.md
alburde1: You should probably remove the dummy content to avoid confusion in that case, since it looks like actual content we need to keep for later. Maybe leave a note in there saying `First technical history release` or something like that.

src/js/helpers/helpHelper.js
```diff
@@ -89,6 +89,25 @@ const loadHistory = () => {
 	return deferred.promise;
 }
 
+const loadTechnicalHistory = () => {
+	const deferred = Q.defer();
+
+	Request.get('/help/history.md')
```
alburde1: if we're loading technical history, shouldn't it be `/help/technicalHistory.md`?

src/js/helpers/helpHelper.js
```diff
@@ -108,6 +127,25 @@ const loadChangelog = () => {
 	return deferred.promise;
 }
 
+const loadTechnicalNotes = () => {
+	const deferred = Q.defer();
+
+	Request.get('/help/technicalHistory.md')
```
alburde1: Should this be `/help/technical.md`?

src/js/containers/router/RouterRoutes.jsx
```diff
@@ -226,6 +226,15 @@ const routeDefinitions = {
             }
         },
         {
+            path: 'technical-history',
```
alburde1: If you look at our other paths, we tend to do camelCase rather than dashes between words. It'd be best to be consistent.

src/js/components/help/historyPage.jsx
```diff
@@ -29,16 +29,29 @@ export default class HelpPage extends React.Component {
...
+                    });
+                })
+                .catch((err) => {
+                    console.log(err);
                 });
-            })
-            .catch((err) => {
-                console.log(err);
-            });
+        }else{
```
alburde1: Just a stylistic thing, we generally do
```
}
else {
```
for if-else blocks. Once again, just being consistent

src/js/components/help/historyPage.jsx
```diff
@@ -20,39 +20,44 @@ export default class HelpPage extends React.Component {
 
         this.state = {
             history: '',
-			sections: []
+			sections: [],
```
alburde1: if we're only using clSections and tSections we don't really need the regular sections state

src/js/components/help/historyPage.jsx
```diff
@@ -20,21 +20,60 @@ export default class HelpPage extends React.Component {
 
         this.state = {
             history: '',
-			sections: []
+            title: '',
+            clSections: [],
+            tSections: [],
+            mounted: false
```
alburde1: This doesn't have to be in the state. It'd be better to just make it a class variable of `isMounted` like we do for other files. Then you wouldn't have to set state while it's unmounting and create potential issues. It also won't interfere with other things and possibly take longer than it needs to to catch that it unmounted and it'll stay consistent with the rest of the components.

Search for "this.isMounted" for examples of how it's used. GenerateDetachedFilesPage.jsx is a very good example of it

src/help/technicalHistory.md
kimminahm: replaced with lorem ipsum

PR #457, opened 2017-02-28
Updating validations downloadable content and table

nmonga91:
Updates to the validation CSVs needed for the deploy


src/help/validations.csv
```diff
@@ -52,33 +53,33 @@ C,C4,"ObligationsDeliveredOrdersUnpaidTotal in File C = USSGL 4901 + 4981 in Fil
...
 D2,D6,"The combination of FAIN, AwardModificationAmendmentNumber, URI, and AwardingSubTierAgencyCode must be unique unless the CorrectionLateDeleteIndicator = C or D.",n/a,Implemented in ASP,Element Conditionally Required
-D2,D7,"This is required only when the combination of FAIN and AwardModificationAmendmentNumber is not unique. Field must contain a unique value except when the CorrectionLateDeleteIndicator contains the code ""C"" or ""D"". This field is required for aggregate records.",n/a,Implemented in ASP,Element Conditionally Required
+D2,D7,This is required only when the combination of FAIN and AwardModificationAmendmentNumber is not unique. Field must contain a unique value except when the CorrectionLateDeleteIndicator contains the code “C” or “D”. This field is required for aggregate records. ,n/a,Implemented in ASP,Element Conditionally Required
```
alburde1: Didn't we have problems last time we changed quotes to smart quotes or am I misremembering? Also, why did we add a space here? There are extra spaces in other changes as well, not going to comment on all of them in case it doesn't matter.

nmonga91: This is a good catch. It was copied & pasted from the CSV I received. I'll fix the quoting.

PR #463, opened 2017-03-06
BUG FIX: Date Selector on detached files was outdated

kimminahm:
fixing offset closing tag, generatedetachedfilespage dateselect had outdated date picker props

### DIFF
--- a/src/js/components/generateDetachedFiles/GenerateDetachedFilesPage.jsx
+++ b/src/js/components/generateDetachedFiles/GenerateDetachedFilesPage.jsx
@@ -158,7 +158,7 @@ export default class GenerateDetachedFilesPage extends React.Component {
         this.setState({[file]:dFile});
     }
 
-    updateError(file, header='', description='') {
+    updateError(file, header, description) {
         // Show any error that occurs at any point during file upload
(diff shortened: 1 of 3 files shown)


src/js/components/generateDetachedFiles/GenerateDetachedFilesPage.jsx
```diff
@@ -158,35 +158,20 @@ export default class GenerateDetachedFilesPage extends React.Component {
...
-				header: '',
-				description: ''
-			}
-		});
-
-		this.setState({
-			[file]: state
-		});
-	}
+    updateError(file, header='', description='') {
```
alburde1: Do we need these defaults if the only function that calls it also has defaults?

kimminahm: Yes we do, this also replaces hideError which will do a call with no parameters.

alburde1: Right, but you pass this down into DateSelect which has its own updateError function with defaults in it. From what I can see, it's the only place it's passed to so it already has the defaults.

PR #466, opened 2017-03-09
Db 1754 mk bug fix context router

kimminahm:
Fixed FABS visual sync. Removed the context.history console warning about deprecated code. Made sure behaviour from validation, back to data entry, back to validation, and any other combination would continue to work

### DIFF
--- a/src/js/components/uploadDetachedFiles/UploadDetachedFileMeta.jsx
+++ b/src/js/components/uploadDetachedFiles/UploadDetachedFileMeta.jsx
@@ -95,7 +95,6 @@ export default class UploadDetachedFileMeta extends React.Component {
 		const start = this.state[file].startDate;
 		const end = this.state[file].endDate;
 
-		// ((start && end) && end.isSameOrAfter(start))
 		dFile.error = {
 				show: ((start && end) && !end.isSameOrAfter(start)) ? true : false,
 				header: ((start && end) && !end.isSameOrAfter(start)) ? 'Invalid Dates' : '',
@@ -141,11 +140,8 @@ export default class UploadDetachedFileMeta extends React.Component {
 
 		this.uploadFileHelper(kGlobalConstants.LOCAL === true && !this.isUnmounted, submission)
 			.then((submissionID) => {
-                    // TODO: Remove this when this is eventually tied to user accounts
                     this.props.setSubmissionId(submissionID);
 					this.checkFileStatus(submissionID);
-					// this.props.history.push('/uploaddetachedfiles/'+submissionID);
-					// this.context.router.transitionTo('/uploadDetachedfiles/'+submissionID);
 					this.props.validate(submissionID);
                 })
 				.catch((err) => {
@@ -158,9 +154,6 @@ export default class UploadDetachedFileMeta extends React.Component {
 
 	checkFileStatus(submissionID) {
 		// callback to check file status
-		if(!submissionID){
-			return;
-		}
 		GenerateFilesHelper.fetchSubmissionMetadata(submissionID)
 			.then((response) => {
 				if (this.isUnmounted) {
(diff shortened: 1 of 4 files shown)


src/js/components/uploadDetachedFiles/UploadDetachedFilesPage.jsx
```diff
@@ -27,24 +29,47 @@ export default class UploadDetachedFilesPage extends React.Component {
 	}
 
 	componentDidMount(){
+		let showMeta = true;
 		if(this.props.params.submissionID){
 			this.props.setSubmissionId(this.props.params.submissionID);
+			showMeta = false;
+		}else{
```
alburde1: This else isn't necessary, it's already true

src/js/components/uploadDetachedFiles/UploadDetachedFilesPage.jsx
```diff
@@ -16,6 +16,8 @@ import UploadDetachedFileValidation from './UploadDetachedFileValidation.jsx';
```
alburde1: Overall question for this file, why are we even bothering with a `showMeta` state? The only time it comes into play other than for setting itself is deciding which to display. Can't you just base it off the params? Or does it not re-render if that's the case?

src/js/components/uploadDetachedFiles/UploadDetachedFilesPage.jsx
```diff
@@ -27,24 +29,47 @@ export default class UploadDetachedFilesPage extends React.Component {
 	}
 
 	componentDidMount(){
+		let showMeta = true;
 		if(this.props.params.submissionID){
```
alburde1: Do we need this part here anymore if we have `componentWillReceiveProps`?

src/js/components/uploadDetachedFiles/UploadDetachedFileMeta.jsx
```diff
@@ -144,8 +144,9 @@ export default class UploadDetachedFileMeta extends React.Component {
                     // TODO: Remove this when this is eventually tied to user accounts
                     this.props.setSubmissionId(submissionID);
 					this.checkFileStatus(submissionID);
-					this.props.history.push('/uploaddetachedfiles/'+submissionID);
-					this.props.validate();
+					// this.props.history.push('/uploaddetachedfiles/'+submissionID);
+					// this.context.router.transitionTo('/uploadDetachedfiles/'+submissionID);
```
kimminahm: done

src/js/components/uploadDetachedFiles/UploadDetachedFileMeta.jsx
```diff
@@ -157,6 +158,9 @@ export default class UploadDetachedFileMeta extends React.Component {
 
 	checkFileStatus(submissionID) {
 		// callback to check file status
+		if(!submissionID){
```
kimminahm: Sorry, got desperate to fix errors and at one point this was receiving submissionID = null. No longer seems to be the case

src/js/components/uploadDetachedFiles/UploadDetachedFileValidation.jsx
```diff
@@ -53,20 +54,30 @@ export default class UploadDetachedFileValidation extends React.Component {
...
+		}
+	}
+
 	componentWillUnmount() {
 		this.isUnmounted = true;
 	}
 
 	checkFileStatus(submissionID) {
 		// callback to check file status
+		if(!submissionID){
```
kimminahm: same as above

src/js/components/uploadDetachedFiles/UploadDetachedFilesPage.jsx
kimminahm: State changes fire off the page to re-render. using only param changes may result in the page refusing to load.

src/js/components/uploadDetachedFiles/UploadDetachedFilesPage.jsx
kimminahm: ComponentWillReceiveProps is only fired off when the props change. ComponentDidMount is necessary for the first setup.

src/js/components/uploadDetachedFiles/UploadDetachedFilesPage.jsx
kimminahm: good catch

src/js/components/uploadDetachedFiles/UploadDetachedFilesPage.jsx
kimminahm: I tried removing all the showMeta stuff and replaced with checking for the props and it didn't work

PR #467, opened 2017-03-09
Db 1632 sortable submissions

kimminahm:
Added sortability to the front end dashboard

### DIFF
--- a/src/js/components/dashboard/DashboardTable.jsx
+++ b/src/js/components/dashboard/DashboardTable.jsx
@@ -75,6 +75,17 @@ export default class DashboardTable extends React.Component {
         buildRow();
     }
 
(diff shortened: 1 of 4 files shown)


src/js/components/dashboard/DashboardTable.jsx
```diff
@@ -66,15 +68,49 @@ export default class DashboardTable extends React.Component {
...
+            case 1:
+                return 'agency';
+                break;
+            case 2:
+                return 'reporting';
+                break;
+            case 3:
+                return 'submitted_by';
+                break;
+            case 4:
```
alburde1: I know it breaks the order but would it make more sense to just remove this case altogether since it does the exact same thing as `default`?

PR #470, opened 2017-03-13
Db 1766 prevent monthly certification

alburde1:
Prevents user from certifying monthly submissions and shows a banner explaining why beneath the buttons.

- [x] Merged after or concurrently with https://github.com/fedspendingtransparency/data-act-broker-backend/pull/610
- [x] Reviewed by frontend

### DIFF
--- a/src/_scss/pages/reviewData/_content.scss
+++ b/src/_scss/pages/reviewData/_content.scss
@@ -8,4 +8,9 @@
     @import "./rightSide/rightSide";
     @import "./bottomSide/bottomSide";
     @import "./narrative/narrative";
+
+    .monthly-submission-error {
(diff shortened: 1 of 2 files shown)


src/_scss/pages/reviewData/_content.scss
```diff
@@ -8,4 +8,9 @@
     @import "./rightSide/rightSide";
     @import "./bottomSide/bottomSide";
     @import "./narrative/narrative";
+
+    .monthly-submission-error {
```
kkuratsu: Please check margin. It has an extra left spacing on smaller screen. Might want to move the error to the same container with certify and notify button`div class="submission-wrapper"`?

PR #481, opened 2017-03-22
Formatting the submission information

kimminahm:
End submission dates should now be the last day of that month on the recent activity table and the tables in the submission dashboard


nmonga91:
This is reformatting dates we get back for a submission, but is the frontend also sending the wrong date range when creating a new submission? In other words, does it send 01/01/2017 - 01/01/2017 for a submission for January 2017?

nmonga91:
New story will be created to handle the dates in the backend.

PR #485, opened 2017-03-23
Removing the message from generateFilesContent, removing LoginTopBar.…

kimminahm:
…jsx and all calls and references because the component is just the error message

removing the component that the bar is, all references to it, and also the warning section on the detached file upload


src/js/components/generateFiles/GenerateFilesContent.jsx
```diff
@@ -35,13 +35,7 @@ export default class GenerateFilesContent extends React.Component {
...
 					</div>
 
-					<div className="usa-da-generate-content">
-                        <div className="alert alert-warning">
-                                <span className="usa-da-icon"><Icons.ExclamationCircle /></span>
-                                <div className="alert-header-text">NOTICE</div>
-                                <p><b>PLEASE NOTE:</b> The D1/D2 integration with ASP and FPDS is still being tested.</p>
-                        </div>
-                        
+					<div className="usa-da-generate-content">
```
alburde1: Tabbing for this whole return (had to put the note here because GitHub)

PR #486, opened 2017-03-23
Adding new button message to check that user is NOT admin and affilia…

kimminahm:
…tion permission is ‘reader’

If a user has a matching affiliation and only has read permissions AND is NOT admin then the button will be blocked with message saying no permissions

### DIFF
--- a/src/js/components/reviewData/ReviewDataContent.jsx
+++ b/src/js/components/reviewData/ReviewDataContent.jsx
@@ -119,6 +119,20 @@ export default class ReviewDataContent extends React.Component {
         return formattedCurrencyString;
     }
 
+    checkAffiliations(){
+        let affiliations = this.props.session.user.affiliations;
+        for(let i = 0; i < affiliations.length; i++){
+            if(affiliations[i].agency_name === this.props.data.agency_name){
+                console.log('match')
+                if(affiliations[i].permission === 'reader'){
+                    return true;
+                }
+                return false;
+            }
+        }
+        return false;
+    }
+
     render() {
 
         let modalToOpen = 'Certify';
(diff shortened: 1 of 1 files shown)


src/js/components/reviewData/ReviewDataContent.jsx
```diff
@@ -119,6 +119,20 @@ export default class ReviewDataContent extends React.Component {
         return formattedCurrencyString;
     }
 
+    checkAffiliations(){
+        let affiliations = this.props.session.user.affiliations;
+        for(let i = 0; i < affiliations.length; i++){
+            if(affiliations[i].agency_name === this.props.data.agency_name){
+                console.log('match')
```
alburde1: Remove the console log

src/js/components/reviewData/ReviewDataContent.jsx
```diff
@@ -180,6 +194,11 @@ export default class ReviewDataContent extends React.Component {
                                         Monthly submissions cannot be certified
                                     </div>
         }
+        if (this.checkAffiliations() && !this.props.session.admin) {
+            certifyButtonText = "You Do Not Have Proper Permissions";
```
alburde1: Is this text cleared with Kaitlin or do we want something more like "You Do Not Have Permission To Certify" or something?

src/js/components/reviewData/ReviewDataContent.jsx
kimminahm: done

src/js/components/reviewData/ReviewDataContent.jsx
kimminahm: You're right. It has been changed

PR #487, opened 2017-03-23
Db 1784 remove a20 a21

alburde1:
Removing rules A20 and A21 from documentation


nmonga91:
Looks good, ready to merge

PR #489, opened 2017-03-27
Submissions are in right table for dashboard and identify subs that need recert

kimminahm:
- [x] needs (db-mk-1793)[https://github.com/fedspendingtransparency/data-act-broker-backend/pull/625] to populate tables correctly and for the recent activity table to render properly

-Updated the recent activity table to mark submissions that need recertification with an updated label
-Updated the dashboard to identify submissions that need recertification

### DIFF
--- a/src/js/components/landing/recentActivity/SubmissionStatus.jsx
+++ b/src/js/components/landing/recentActivity/SubmissionStatus.jsx
@@ -63,11 +63,15 @@ export class SubmissionStatus extends React.Component {
 	render() {
 
 		const colors = this.progressBar(this.props.status);
+		let label = this.statusStrings[this.props.status]
+		if(this.props.status !== 5 && this.props.certified){
+			label += '\n(Needs Recertification)'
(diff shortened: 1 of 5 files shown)


src/js/components/landing/recentActivity/SubmissionStatus.jsx
```diff
@@ -63,11 +63,15 @@ export class SubmissionStatus extends React.Component {
 	render() {
 
 		const colors = this.progressBar(this.props.status);
+		let label = this.statusStrings[this.props.status]
```
alburde1: Did you check what the wording should be for the status on these? I seem to remember discussing how it should just say "Updated" but we should check.

kimminahm: Double Checked with Kaitlin. Since they can have different status besides update (validating, validated, validated with errors, etc.) it was better to just add (need recert) after the fact

PR #495, opened 2017-03-29
Update Changelog and Archive old notes

kimminahm:
Formatted and added new changelog and moved current release notes into archive


nmonga91:
I see the change log updated here but not the technical release notes, is that intentional?

PR #537, opened 2017-05-01
Db 1547 mk navbar permissions

kimminahm:
Using user permissions to render navbar and update the first block on the landing page


nmonga91:
For the 2nd acceptance criteria, the text within the box for Upload and Validate should also change. Right now it says "Ready to start uploading and validating your agency's files? Click here to get setup.*" and "What you will need to submit". We may want to consider updating that to say something like "In order to upload and validate your agency's files, please follow the link below to request access". What do you think @kaitlin ?

src/js/components/SharedComponents/navigation/NavigationComponent.jsx
```diff
@@ -32,21 +32,47 @@ export class Navbar extends React.Component {
...
 
-    render() {
-        let tabNames = {}
+    checkPermissions(){
+        if(!this.props.session.user.affiliations || this.props.session.user.affiliations.length == 0){
+            return false;
+        }
+        let aff = this.props.session.user.affiliations;
+        for(let i = 0; i < aff.length; i++){
+            if(aff[i].permission !== 'reader'){
```
nmonga91: Is there a reason for using `!==` vs `!=` here? the "===" is strict and makes sure the data types are also the same, is that necessary here?

src/js/components/dashboard/DashboardContent.jsx
```diff
@@ -10,6 +10,7 @@ export default class DashboardContent extends React.Component {
     constructor(props) {
         super(props);
 
+        console.log(this.props)
```
nmonga91: This should probably be removed

src/js/components/dashboard/DashboardTable.jsx
```diff
@@ -32,6 +32,9 @@ const tableHeaders = [
 export default class DashboardTable extends React.Component {
     constructor(props) {
         super(props);
+
+        console.log(this.props)
```
nmonga91: Should remove this console.log

src/js/components/landing/LandingContent.jsx
```diff
@@ -61,6 +61,22 @@ export default class LandingContent extends React.Component {
         })
     }
 
+    checkPermissions(){
```
nmonga91: This is the identical to the `checkPermissions()` in NavigationComponent.jsx. Instead of duplicating it, it should be moved a helper and then reused that way.

src/js/components/SharedComponents/navigation/NavigationComponent.jsx
kimminahm: Bray said to just use strict typing wherever
I can change it back though np

src/js/components/landing/LandingContent.jsx
kimminahm: Thats a good point. Will do

src/js/components/SharedComponents/navigation/NavigationComponent.jsx
nmonga91: If that was the suggestion, we should stick with that.

nmonga91:
Looks good, ready to merge

PR #544, opened 2017-05-10
DB-0000-Wording-Changes

kimminahm:
Wording Changes based of e-mail thread. Can Forward if needed

### DIFF
--- a/src/js/components/landing/LandingContent.jsx
+++ b/src/js/components/landing/LandingContent.jsx
@@ -79,7 +79,7 @@ export default class LandingContent extends React.Component {
                                 <div className="col-md-7 mt-40 mb-50">
                                     <h1 className="display-2" data-contentstart="start" tabIndex={-1}>Welcome to the DATA Act Broker</h1>
                                     <p>Upload your agency’s files and validate them against the latest version of the DATA Act Information Model Schema (DAIMS).</p>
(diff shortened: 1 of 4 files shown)


src/js/components/landing/LandingContent.jsx
```diff
@@ -103,43 +78,39 @@ export default class LandingContent extends React.Component {
...
                                 <div className="col-md-7 mt-40 mb-50">
                                     <h1 className="display-2" data-contentstart="start" tabIndex={-1}>Welcome to the DATA Act Broker</h1>
-                                    <p>Upload your agency's files and validate them against the latest version of the DATA Act Information Model Schema.</p>
-                                    <p>Details on how to format your data, including required and optional fields, can be found in the <a href={this.state.rssUrl} target="_blank" rel="noopener noreferrer" aria-label="Download the RSS specification as an Excel file">Reporting Submission Specification (RSS) v1.0</a>.</p>
+                                    <p>Upload your agency’s files and validate them against the latest version of the DATA Act Information Model Schema (DAIMS).</p>
+                                    <p>Details on how to format your data, including required and optional fields, can be found in the <a href="https://broker.usaspending.gov/#/resources?_k=4bsu0y" target="_blank" rel="noopener noreferrer" >Resources section.</a>.</p>
```
alburde1: we probably want to leave off the `?_k=4bsu0y` portion of that link because that's basically to avoid cache issues and such. It'll auto-add one once they get there if you just use `https://broker.usaspending.gov/#/resources`

kimminahm: fixed and thanks for pointing that out

PR #551, opened 2017-05-15
db-1700 Add the rule label for error and warning descriptions

kkuratsu:
As an agency user, I want to see the rule label in all of the error and warning descriptions

**Acceptance Criteria:**

- Rule label (i.e. A9) is present in the error messages on all validation pages

**Checklist:**
- [ ] Backend changes: https://github.com/fedspendingtransparency/data-act-broker-backend/pull/699

### DIFF
--- a/src/js/components/validateData/validateValues/ValidateValuesErrorReport.jsx
+++ b/src/js/components/validateData/validateValues/ValidateValuesErrorReport.jsx
@@ -66,11 +66,7 @@ export default class ValidateValuesErrorReport extends React.Component {
 
             let description = item.error_description;
             if (item.error_name == 'rule_failed') {
-                // Remove "Failed rule:" label
-                let msg = item.rule_failed.split('Failed rule: ');
-                msg.shift();
-                msg = msg.join('Failed rule: ');
-                description = ['Rule ' + item.original_label + ': ' + msg];
+                description = ['Rule ' + item.original_label + ': ' + item.rule_failed];
             }
 
             const row = [item.field_name, description, parseInt(item.occurrences)];
--- a/src/js/components/validateData/validateValues/ValidateValuesErrorReport.jsx
+++ b/src/js/components/validateData/validateValues/ValidateValuesErrorReport.jsx
@@ -66,7 +66,7 @@ export default class ValidateValuesErrorReport extends React.Component {
 
             let description = item.error_description;
             if (item.error_name == 'rule_failed') {
-                description = ['Rule ' + item.original_label + ': ' + item.rule_failed];
+                description = 'Rule ' + item.original_label + ': ' + item.rule_failed;
             }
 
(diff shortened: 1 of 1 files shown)


src/js/components/validateData/validateValues/ValidateValuesErrorReport.jsx
```diff
@@ -66,7 +66,7 @@ export default class ValidateValuesErrorReport extends React.Component {
 
             let description = item.error_description;
             if (item.error_name == 'rule_failed') {
-                description = item.rule_failed;
+                description = ['Rule ' + item.original_label + ': ' + item.rule_failed];
```
alburde1: It seems redundant to keep `Failed rule:` in the display. If you look at the cross-file validations, it just says `Rule LABEL: DESCRIPTION`. I think that's a backend change but I think we should keep it consistent

kkuratsu: Please look at the next line (https://github.com/fedspendingtransparency/data-act-broker-web-app/blob/665a2a4c34f9aadb2b9d2872cac5a11cb6c005c4/src/js/components/validateData/validateValues/ValidateValuesErrorReport.jsx#L68-L72). 

I followed example of the cross-file validation (https://github.com/fedspendingtransparency/data-act-broker-web-app/blob/dev/src/js/components/crossFile/components/ComparisonTable.jsx#L27)

alburde1: Right, but the actual text that is returned in `item.rule_failed` is different. Because we didn't have labels in this one originally it said `Failed rule: RULE TEXT` so now, with the change, it says `Rule LABEL: Failed rule: RULE TEXT`.

It seems redundant especially since in cross-file validations it just says `Rule LABEL: RULE TEXT`

kkuratsu:
I changed the front-end to remove `Failed rule:`. Let me know if that’s ok with you. I am not quite certain if remove `Failed rule:` from the backend will be a good idea, there are some other places using that `item.rule_failed`.

alburde1:
The only other location, I believe, is the treemap, which currently says
"More information:
Failed rule: RULE DESCRIPTION"
Which I don't think REALLY needs the "Failed rule" part considering it's kind of implied when it's in the treemap.

kkuratsu:
I changed the backend to output `rule_description` instead of `rule_error_message` as suggested. I notice that those two columns are pretty similar with only `Failed rule: ` preceding the descriptions in the `rule_error_message` column. Thank you!

src/js/components/validateData/validateValues/ValidateValuesErrorReport.jsx
alburde1: As a side note, I don't think you want to put it into an array. You can just make it a string

PR #554, opened 2017-05-15
Db 1777 mk page navigation

kimminahm:
- [x] Frontend Review
- [x] Dependent on backend [branch](https://github.com/fedspendingtransparency/data-act-broker-backend/pull/691)

Allows for navigation to most current step from all tables, allows for people to go directly to the stage they are at and prevents people from navigating too far

### DIFF
--- a/src/js/components/landing/recentActivity/SubmissionLink.jsx
+++ b/src/js/components/landing/recentActivity/SubmissionLink.jsx
@@ -5,6 +5,7 @@
 
 import React from 'react';
 import * as Icons from '../../SharedComponents/icons/Icons.jsx';
+import * as SubmissionHelper from '../../../helpers/submissionGuideHelper.js';
 
 const propTypes = {
 	submissionId: React.PropTypes.oneOfType([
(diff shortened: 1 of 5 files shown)


src/js/containers/generateFiles/GenerateFilesContainer.jsx
```diff
@@ -126,9 +126,7 @@ class GenerateFilesContainer extends React.Component {
 					}
 					// certified submission files have been updated, hide prepopulate D1/D2
```
nmonga91: There is a merge conflict which is probably caused by the updates to this happening in a separate branch (which has already been merged). You will need to merge dev back into your branch to resolve this.

src/js/components/landing/recentActivity/SubmissionLink.jsx
```diff
@@ -18,10 +19,25 @@ const defaultProps = {
 }
 
 export default class SubmissionLink extends React.Component {
+	constructor(props) {
+		super(props);
+		this.state = {
+			page: '#/'
```
nmonga91: Can the page ever be redirected to the home page? If not, we can just have this preset to an empty string instead.

src/js/helpers/submissionGuideHelper.js
```diff
@@ -29,4 +29,37 @@ export const setSkipGuide = (skip_guide) => {
...
+    const store = new StoreSingleton().store;
+
+    Request.get(kGlobalConstants.API + 'check_current_page/?submission_id=' + submission_id)
+        .end((err, res) => {
+            if (err) {
+                deferred.reject(err);
+            } else {
+                // Only skip the guide if the user wants to skip the guide
+                let pages = [
+                    '/404',
```
nmonga91: So if it's not one of the other pages, does it default to the 404 page?

kimminahm: 0 returns if the submission files dont exist. Kaitlin said that that it should be fine to just redirect to a 404 page when no content exists
screen shot 2017-05-16 at 4 03 34 pm

src/js/components/landing/recentActivity/SubmissionLink.jsx
kimminahm: Good point, empty string works, actually i think it may be better to go to a 404 because the only time that will happen is when the api returns a '0' which means the submission files do not exist

nmonga91:
Looks good, merging

PR #567, opened 2017-05-24
Upgrade package management to yarn

EricSchles:
This was the only way I could install all the node packages.  `npm install` literally doesn't work.  Here are some resources on why I think we should use `yarn` for the web app moving forward:
* https://code.facebook.com/posts/1840075619545360 - an introduction to the yarn package manager
* We are already using it's predecessor - `npm-shrinkwrap.json` - yarn is under more active development than `npm-shrinkwrap.json` and has better support
* `yarn` worked where `npm install` failed (at least for me).
* emoji's in your install behavior - Among other things, yarn shows you it's progress with emoji's and what could be better than that?!
screen shot 2017-05-24 at 8 07 19 am


klrBAH:
This will be addressed via correcting node version and a library update in a separate user story

PR #575, opened 2017-05-26
DB 1901 Generate Detached D files

kimminahm:
- [x] Backend First https://github.com/fedspendingtransparency/data-act-broker-backend/pull/715
- [ ] Front End Review

ability to generate D files seperately


src/js/containers/SharedContainers/AgencyListContainer.jsx
```diff
@@ -21,7 +21,7 @@ class AgencyListContainer extends React.Component {
 	loadData() {
 		if (this.props.agencyList.agencies.length == 0) {
 			// we need to populate the list
-			AgencyHelper.fetchAgencies()
+			AgencyHelper.fetchAllAgencies()
```
alburde1: My concern here is that this will also fetch all agencies for the regular submission agency list, which is incorrect. It should only fetch all for the detached D file generation.

PR #587, opened 2017-06-07
Hotfix new submission 404

kimminahm:
Prevents the 404 when new submissions are created

### DIFF
--- a/src/js/containers/addData/AddDataContainer.jsx
+++ b/src/js/containers/addData/AddDataContainer.jsx
@@ -48,7 +48,7 @@ class AddDataContainer extends React.Component {
 					for(let i = 0; i < count; i++){
 						GuideHelper.getSubmissionPage(submissionID)
 							.then((res) => {
-								hashHistory.push('/validateData/' + submissionID)
+								hashHistory.push('/validateData/' + submissionID);
 							})
 							.catch((err) => {
 								if(i == count - 1) {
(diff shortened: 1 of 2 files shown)


src/js/containers/addData/AddDataContainer.jsx
```diff
@@ -41,27 +42,62 @@ class AddDataContainer extends React.Component {
...
-
+					// Looping because we need to allow backend to catchup to front end and prevent incorrect 404
+					let count = 10;
+					for(let i = 0; i < count; i++){
+						GuideHelper.getSubmissionPage(submissionID)
+							.then((res) => {
+								hashHistory.push('/validateData/' + submissionID);
+							})
+							.catch((err) => {
+								if(i == count - 1) {
```
alburde1: Not that it's a huge deal, but if we want to limit it to 10 we can either start i at 1 or set count to 9 rather than doing this calculation every time. It's not an intensive calculation, but why do any extra processing when we don't need to?

src/js/containers/addData/AddDataContainer.jsx
```diff
@@ -41,27 +42,62 @@ class AddDataContainer extends React.Component {
...
+					this.props.setSubmissionId(submissionID);
+					// Looping because we need to allow backend to catchup to front end and prevent incorrect 404
+					let count = 10;
+					for(let i = 0; i < count; i++){
+						GuideHelper.getSubmissionPage(submissionID)
+							.then((res) => {
+								hashHistory.push('/validateData/' + submissionID);
+							})
+							.catch((err) => {
+								if(i == count - 1) {
```
alburde1: Same comment as above for count.

PR #607, opened 2017-06-15
Hotfix multiagency login hanging

klrBAH:


src/_scss/pages/landing/recentActivity/_header.scss
```diff
@@ -9,7 +9,6 @@
     font-weight: 200;
     line-height: 1;
     overflow: hidden;
-    height: 1em;
```
alburde1: Just the one question, why are we removing this height?

klrBAH: Hard setting the height restricts it so that you can't see the full truncated text of agency names in the list, you only get one line even if the list flows to more than one line.

PR #627, opened 2017-07-03
DB-mk-1968 daims resources

kimminahm:
Notification to users regarding daims and broker v1.1 along with links to the max documentation
Also fixed indets


dpb-bah:
Looks like the DAIMS message is on the required pages and code looks fine, merging

PR #637, opened 2017-07-10
Db mk 1944 gtas window restriction

kimminahm:
- [x] Required backend https://github.com/fedspendingtransparency/data-act-broker-backend/pull/774

### DIFF
--- a/src/js/components/generateEF/GenerateEFPage.jsx
+++ b/src/js/components/generateEF/GenerateEFPage.jsx
@@ -13,8 +13,6 @@ import * as ReviewHelper from '../../helpers/reviewHelper.js';
 import GenerateEFContainer from '../../containers/generateEF/GenerateEFContainer.jsx';
 import GenerateEFError from './GenerateEFError.jsx';
 
-import GTASBanner from '../SharedComponents/GTASWarningBanner.jsx';
-
 export default class GenerateEFPage extends React.Component {
 
     constructor(props) {
@@ -22,13 +20,8 @@ export default class GenerateEFPage extends React.Component {
 
         this.state = {
             showError: false,
-            errorMessage: '',
-            gtas: null
+            errorMessage: ''
         };
-    }
-
-    componentDidMount() {
-        this.isGtas();
     }
 
     showError(message) {
@@ -38,29 +31,12 @@ export default class GenerateEFPage extends React.Component {
         });
     }
 
-    isGtas() {
-        ReviewHelper.isGtasWindow()
-            .then((res) => {
-                if(res != this.state.gtas) {
-                    this.setState({gtas: res})
-                }
-            })
-            .catch((err) =>{
-                console.log(err)
-            })
-    }
-
 	render() {
 
         let pageContent = <GenerateEFContainer submissionID={this.props.params.submissionID} showError={this.showError.bind(this)} />;
 
         if (this.state.showError) {
            pageContent = <GenerateEFError message={this.state.errorMessage} />;
-        }
-
-        let gtasWarning = null;
-        if(this.state.gtas && this.state.gtas.data) {
-            gtasWarning = <GTASBanner data={this.state.gtas}/>
         }
 
 		return (
@@ -74,7 +50,6 @@ export default class GenerateEFPage extends React.Component {
                         </div>
                     </div>
                 </div>
-                {gtasWarning}
                 {pageContent}
             </div>
             
--- a/src/js/components/generateEF/GenerateEFPage.jsx
+++ b/src/js/components/generateEF/GenerateEFPage.jsx
@@ -8,8 +8,6 @@ import Navbar from '../SharedComponents/navigation/NavigationComponent.jsx';
 import AddDataHeader from './../addData/AddDataHeader.jsx';
 import Progress from '../SharedComponents/ProgressComponent.jsx';
 
-import * as ReviewHelper from '../../helpers/reviewHelper.js';
-
 import GenerateEFContainer from '../../containers/generateEF/GenerateEFContainer.jsx';
 import GenerateEFError from './GenerateEFError.jsx';
 
--- a/src/js/components/generateEF/GenerateEFPage.jsx
+++ b/src/js/components/generateEF/GenerateEFPage.jsx
@@ -8,8 +8,12 @@ import Navbar from '../SharedComponents/navigation/NavigationComponent.jsx';
 import AddDataHeader from './../addData/AddDataHeader.jsx';
 import Progress from '../SharedComponents/ProgressComponent.jsx';
 
+import * as ReviewHelper from '../../helpers/reviewHelper.js';
+
 import GenerateEFContainer from '../../containers/generateEF/GenerateEFContainer.jsx';
 import GenerateEFError from './GenerateEFError.jsx';
+
+import GTASBanner from '../SharedComponents/GTASWarningBanner.jsx';
 
 export default class GenerateEFPage extends React.Component {
 
@@ -18,8 +22,13 @@ export default class GenerateEFPage extends React.Component {
 
         this.state = {
             showError: false,
-            errorMessage: ''
+            errorMessage: '',
+            gtas: null
         };
+    }
+
+    componentDidMount() {
+        this.isGtas();
     }
 
     showError(message) {
@@ -29,12 +38,29 @@ export default class GenerateEFPage extends React.Component {
         });
     }
 
+    isGtas() {
+        ReviewHelper.isGtasWindow()
+            .then((res) => {
+                if(res != this.state.gtas) {
+                    this.setState({gtas: res})
+                }
+            })
+            .catch((err) =>{
+                console.log(err)
+            })
+    }
+
 	render() {
 
         let pageContent = <GenerateEFContainer submissionID={this.props.params.submissionID} showError={this.showError.bind(this)} />;
 
         if (this.state.showError) {
            pageContent = <GenerateEFError message={this.state.errorMessage} />;
+        }
+
+        let gtasWarning = null;
+        if(this.state.gtas && this.state.gtas.data) {
+            gtasWarning = <GTASBanner data={this.state.gtas}/>
         }
 
 		return (
@@ -48,7 +74,7 @@ export default class GenerateEFPage extends React.Component {
                         </div>
                     </div>
                 </div>
-
+                {gtasWarning}
                 {pageContent}
             </div>
             
--- a/src/js/components/reviewData/ReviewDataContent.jsx
+++ b/src/js/components/reviewData/ReviewDataContent.jsx
@@ -175,17 +175,21 @@ export default class ReviewDataContent extends React.Component {
 
         let certifyButtonText = "You do not have permissions to certify";
         let notifyButtonText = "Notify Another User that the Submission is Ready for Certification";
-        let buttonClass = "btn-disabled";
+        let buttonClass = " btn-disabled";
         let buttonAction = "";
         let monthlySubmissionError = null;
 
-        if (this.props.data.publish_status == "published") {
+        if(this.props.data.gtas && this.props.data.gtas) {
+            certifyButtonText = "Certification is not allowed during the GTAS Submission Window";
+            buttonClass = " btn-disabled";
+            buttonAction = "";
+        }
+        else if (this.props.data.publish_status == "published") {
             certifyButtonText = "Submission has already been certified";
             buttonClass = " btn-disabled";
             buttonAction = "";
         }
-
-        if (!this.props.data.quarterly_submission) {
+        else if (!this.props.data.quarterly_submission) {
             certifyButtonText = "Monthly submissions cannot be certified";
             buttonClass = " btn-disabled";
             buttonAction = "";
@@ -194,18 +198,13 @@ export default class ReviewDataContent extends React.Component {
                                         Monthly submissions cannot be certified
                                     </div>
         }
-
-        if ((this.checkAffiliations() || this.props.session.admin ) && buttonClass != " btn-disabled") {
+        else if (this.checkAffiliations() || this.props.session.admin) {
             certifyButtonText = "Certify & Publish the Submission to USAspending.gov";
             buttonClass = "";
             buttonAction = this.openModal.bind(this, modalToOpen);
         }
 
-        if(buttonClass != " btn-disabled" && this.props.data.gtas && this.props.data.gtas.data) {
-            certifyButtonText = "Certification is not allowed during the GTAS Submission Window";
-            buttonClass = " btn-disabled";
-            buttonAction = "";
-        }
+        
 
         return (
             <div className="container">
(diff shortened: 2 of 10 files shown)


src/js/components/reviewData/ReviewDataContent.jsx
```diff
@@ -192,10 +194,17 @@ export default class ReviewDataContent extends React.Component {
...
-        if (this.checkAffiliations() || this.props.session.admin) {
+
+        if ((this.checkAffiliations() || this.props.session.admin ) && buttonClass != " btn-disabled") {
             certifyButtonText = "Certify & Publish the Submission to USAspending.gov";
             buttonClass = "";
-            buttonAction = this.openModal.bind(this, modalToOpen);;
+            buttonAction = this.openModal.bind(this, modalToOpen);
+        }
+
+        if(buttonClass != " btn-disabled" && this.props.data.gtas.open) {
```
alburde1: Based on changes to the backend, gtas.open doesn't exist anymore.

src/js/containers/crossFile/CrossFileContentContainer.jsx
```diff
@@ -42,6 +47,18 @@ class CrossFileContentContainer extends React.Component {
 		}
 	}
 
+	isGtas() {
+		ReviewHelper.isGtasWindow()
+			.then((res) => {
+				if(res.open != this.state.gtas) {
+					this.setState({gtas: res})
```
alburde1: Since all we're passing back now is the "data" element in the JSON (which is either `null` or an object) it might make sense to set gtas to `res.data`)

src/js/containers/generateEF/GenerateEFContainer.jsx
```diff
@@ -33,14 +35,28 @@ class GenerateEFContainer extends React.Component {
...
 		this.isUnmounted = false;
 		this.checkFileStatus();
+		this.isGtas();
+	}
+
+	isGtas() {
+		ReviewHelper.isGtasWindow()
+			.then((res) => {
+				if(res.open != this.state.gtas) {
+					this.setState({gtas: res})
```
alburde1: Since all we're passing back now is the "data" element in the JSON (which is either `null` or an object) it might make sense to set gtas to `res.data`)

src/js/containers/generateFiles/GenerateFilesContainer.jsx
```diff
@@ -60,15 +62,30 @@ class GenerateFilesContainer extends React.Component {
...
 		this.isUnmounted = false;
 		this.checkForPreviousFiles();
+		this.isGtas();
 	}
 
+	isGtas() {
+		ReviewHelper.isGtasWindow()
+			.then((res) => {
+				if(res.open != this.state.gtas) {
+					this.setState({gtas: res})
```
alburde1: Since all we're passing back now is the "data" element in the JSON (which is either `null` or an object) it might make sense to set gtas to `res.data`)

src/js/containers/review/ReviewDataContainer.jsx
```diff
@@ -43,6 +45,18 @@ class ReviewDataContainer extends React.Component {
     	}
     }
 
+    isGtas() {
+        ReviewHelper.isGtasWindow()
+            .then((res) => {
+                if(res.open != this.state.gtas) {
+                    this.setState({gtas: res})
```
alburde1: Since all we're passing back now is the "data" element in the JSON (which is either `null` or an object) it might make sense to set gtas to `res.data`)

src/js/containers/validateData/ValidateDataContainer.jsx
```diff
@@ -60,6 +63,18 @@ class ValidateDataContainer extends React.Component {
 		}
 	}
 
+	isGtas() {
+		ReviewHelper.isGtasWindow()
+			.then((res) => {
+				if(res.open != this.state.gtas) {
+					this.setState({gtas: res})
```
alburde1: Since all we're passing back now is the "data" element in the JSON (which is either `null` or an object) it might make sense to set gtas to `res.data`)

src/js/components/reviewData/ReviewDataPage.jsx
```diff
@@ -30,6 +31,11 @@ export default class ReviewDataPage extends React.Component {
             warningMessage = <PublishedSubmissionWarningBanner />;
         }
 
+        let gtasWarning = null;
+        if(this.props.data.gtas && this.props.data.gtas.open) {
```
alburde1: `gtas.open` no longer exists. Given the suggestion below you can probably just do the first part of this `if` for checking.

src/js/containers/crossFile/CrossFileContentContainer.jsx
```diff
@@ -214,9 +231,16 @@ class CrossFileContentContainer extends React.Component {
 		if(this.props.submission.publishStatus !== "unpublished") {
 			warningMessage = <PublishedSubmissionWarningBanner />;
 		}
+
+		let gtasWarning = null;
+		if(this.state.gtas && this.state.gtas.open){
```
alburde1: gtas.open no longer exists. Given the suggestion below you can probably just do the first part of this if for checking.

src/js/containers/generateEF/GenerateEFContainer.jsx
```diff
@@ -154,10 +170,16 @@ class GenerateEFContainer extends React.Component {
 		if(this.props.submission.publishStatus !== "unpublished") {
 			warningMessage = <PublishedSubmissionWarningBanner />;
 		}
+
+		let gtasWarning = null;
+		if(this.state.gtas && this.state.gtas.open){
```
alburde1: gtas.open no longer exists. Given the suggestion below you can probably just do the first part of this if for checking.

src/js/containers/generateFiles/GenerateFilesContainer.jsx
```diff
@@ -469,9 +486,15 @@ class GenerateFilesContainer extends React.Component {
 		if(this.props.submission.publishStatus !== "unpublished") {
 			warningMessage = <PublishedSubmissionWarningBanner />;
 		}
+
+		let gtasWarning = null;
+		if(this.state.gtas && this.state.gtas.open){
```
alburde1: gtas.open no longer exists. Given the suggestion below you can probably just do the first part of this if for checking.

src/js/containers/validateData/ValidateDataContainer.jsx
```diff
@@ -200,9 +215,15 @@ class ValidateDataContainer extends React.Component {
 			warningMessage = <PublishedSubmissionWarningBanner />;
 		}
 
+		let gtasWarning = null;
+		if(this.state.gtas && this.state.gtas.open){
```
alburde1: gtas.open no longer exists. Given the suggestion below you can probably just do the first part of this if for checking.

src/js/containers/crossFile/CrossFileContentContainer.jsx
```diff
@@ -42,6 +47,18 @@ class CrossFileContentContainer extends React.Component {
 		}
 	}
 
+	isGtas() {
+		ReviewHelper.isGtasWindow()
+			.then((res) => {
+				if(res.open != this.state.gtas) {
```
alburde1: Even in the previous setup, `res.open` would NEVER equal `this.state.gtas`. Given the suggestion below and the changes to what we return from the backend, this check could be removed completely and you could always just set state.

src/js/containers/generateEF/GenerateEFContainer.jsx
```diff
@@ -33,14 +35,28 @@ class GenerateEFContainer extends React.Component {
...
 	componentDidMount() {
 		this.isUnmounted = false;
 		this.checkFileStatus();
+		this.isGtas();
+	}
+
+	isGtas() {
+		ReviewHelper.isGtasWindow()
+			.then((res) => {
+				if(res.open != this.state.gtas) {
```
alburde1: Even in the previous setup, res.open would NEVER equal this.state.gtas. Given the suggestion below and the changes to what we return from the backend, this check could be removed completely and you could always just set state.

src/js/containers/generateFiles/GenerateFilesContainer.jsx
```diff
@@ -60,15 +62,30 @@ class GenerateFilesContainer extends React.Component {
...
 	componentDidMount() {
 		this.isUnmounted = false;
 		this.checkForPreviousFiles();
+		this.isGtas();
 	}
 
+	isGtas() {
+		ReviewHelper.isGtasWindow()
+			.then((res) => {
+				if(res.open != this.state.gtas) {
```
alburde1: Even in the previous setup, res.open would NEVER equal this.state.gtas. Given the suggestion below and the changes to what we return from the backend, this check could be removed completely and you could always just set state.

src/js/containers/review/ReviewDataContainer.jsx
```diff
@@ -43,6 +45,18 @@ class ReviewDataContainer extends React.Component {
     	}
     }
 
+    isGtas() {
+        ReviewHelper.isGtasWindow()
+            .then((res) => {
+                if(res.open != this.state.gtas) {
```
alburde1: Even in the previous setup, res.open would NEVER equal this.state.gtas. Given the suggestion below and the changes to what we return from the backend, this check could be removed completely and you could always just set state.

src/js/containers/validateData/ValidateDataContainer.jsx
```diff
@@ -60,6 +63,18 @@ class ValidateDataContainer extends React.Component {
 		}
 	}
 
+	isGtas() {
+		ReviewHelper.isGtasWindow()
+			.then((res) => {
+				if(res.open != this.state.gtas) {
```
alburde1: Even in the previous setup, res.open would NEVER equal this.state.gtas. Given the suggestion below and the changes to what we return from the backend, this check could be removed completely and you could always just set state.

src/js/components/generateEF/GenerateEFPage.jsx
```diff
@@ -8,27 +8,48 @@ import Navbar from '../SharedComponents/navigation/NavigationComponent.jsx';
...
             showError: true,
             errorMessage: message
         });
     }
 
+    isGtas() {
+        ReviewHelper.isGtasWindow()
+            .then((res) => {
+                if(res != this.state.gtas) {
+                    this.setState({gtas: res})
```
alburde1: We can still just set gtas to `res.data` here. Apply to all below. There is no reason not to because you're always passing "data" back, it is just either None or contains an object.

src/js/components/generateEF/GenerateEFPage.jsx
```diff
@@ -8,6 +8,8 @@ import Navbar from '../SharedComponents/navigation/NavigationComponent.jsx';
 import AddDataHeader from './../addData/AddDataHeader.jsx';
 import Progress from '../SharedComponents/ProgressComponent.jsx';
 
+import * as ReviewHelper from '../../helpers/reviewHelper.js';
```
alburde1: Why is this being passed in and not used? I assume you meant to add the GTAS banner to this page too? If not, this should be removed.

src/js/components/reviewData/ReviewDataContent.jsx
```diff
@@ -192,10 +194,17 @@ export default class ReviewDataContent extends React.Component {
...
-        if (this.checkAffiliations() || this.props.session.admin) {
+
+        if ((this.checkAffiliations() || this.props.session.admin ) && buttonClass != " btn-disabled") {
             certifyButtonText = "Certify & Publish the Submission to USAspending.gov";
             buttonClass = "";
-            buttonAction = this.openModal.bind(this, modalToOpen);;
+            buttonAction = this.openModal.bind(this, modalToOpen);
+        }
+
+        if(buttonClass != " btn-disabled" && this.props.data.gtas && this.props.data.gtas.data) {
```
alburde1: If you're already setting gtas to `res.data` you don't have to check for `gtas.data` (it'll always be empty actually because it doesn't exist).

src/js/components/reviewData/ReviewDataPage.jsx
```diff
@@ -30,6 +31,11 @@ export default class ReviewDataPage extends React.Component {
             warningMessage = <PublishedSubmissionWarningBanner />;
         }
 
+        let gtasWarning = null;
+        if(this.props.data.gtas && this.props.data.gtas) {
```
alburde1: You're checking the same thing twice

src/js/components/reviewData/ReviewDataContent.jsx
```diff
@@ -192,10 +194,17 @@ export default class ReviewDataContent extends React.Component {
                                         Monthly submissions cannot be certified
                                     </div>
         }
-        if (this.checkAffiliations() || this.props.session.admin) {
+
+        if ((this.checkAffiliations() || this.props.session.admin ) && buttonClass != " btn-disabled") {
```
alburde1: Should all of these "if" statements be "else if"? I'm not sure, go through and check because you may be able to simplify them if they're else if

src/js/components/generateEF/GenerateEFPage.jsx
kimminahm: it was hear but i ended up adding it to the container so i removed it

src/js/components/reviewData/ReviewDataContent.jsx
```diff
@@ -175,15 +175,21 @@ export default class ReviewDataContent extends React.Component {
 
         let certifyButtonText = "You do not have permissions to certify";
         let notifyButtonText = "Notify Another User that the Submission is Ready for Certification";
-        let buttonClass = "btn-disabled";
+        let buttonClass = " btn-disabled";
         let buttonAction = "";
         let monthlySubmissionError = null;
-        if (this.props.data.publish_status == "published") {
+
+        if(this.props.data.gtas && this.props.data.gtas) {
```
alburde1: Do we want to display this over "already certified" and "monthly cannot be certified"? I'd imagine we'd want to display those first, then "certification not allowed during GTAS" only if neither of those is true. I think those messages are more important

src/js/components/reviewData/ReviewDataContent.jsx
```diff
@@ -192,12 +193,19 @@ export default class ReviewDataContent extends React.Component {
                                         Monthly submissions cannot be certified
                                     </div>
         }
-        if (this.checkAffiliations() || this.props.session.admin) {
+        else if(this.props.data.gtas && this.props.data.gtas) {
```
alburde1: Still checking the same thing twice here.

src/js/components/reviewData/ReviewDataContent.jsx
```diff
@@ -175,15 +175,16 @@ export default class ReviewDataContent extends React.Component {
...
         let certifyButtonText = "You do not have permissions to certify";
         let notifyButtonText = "Notify Another User that the Submission is Ready for Certification";
-        let buttonClass = "btn-disabled";
+        let buttonClass = " btn-disabled";
         let buttonAction = "";
         let monthlySubmissionError = null;
+
         if (this.props.data.publish_status == "published") {
             certifyButtonText = "Submission has already been certified";
             buttonClass = " btn-disabled";
```
alburde1: In all of these "if" statements other than the last one, you don't have to change the buttonAction or buttonClass because you're just changing it to what's already there. The only time it has to be changed is if we stop it from being disabled and allow the user to click on it.

PR #676, opened 2017-07-28
DB 2102 FABS has readonly access to DABS

michaeldhess:
AC:
- Users with F permissions have read-only access to DABS

### DIFF
--- a/src/js/components/landing/BlockContent.jsx
+++ b/src/js/components/landing/BlockContent.jsx
@@ -22,21 +22,8 @@ export default class BlockContent extends React.Component {
         let thirdBlock = null;
 
         if (this.props.type === 'home') {
-            if (this.props.session.user.helpOnly) {
-                // No access
-                firstBlock = <LandingBlock type={this.props.type} icon={<h5>DATA Act Broker Submission</h5>} text="In order to view the DATA Act Broker Submissions, please follow the link below to request access" buttonText="Request Access" url="https://community.max.gov/x/fJwuRQ" />;
-                secondBlock = <LandingBlock type={this.props.type} icon={<h5>Financial Assistance Broker Submission</h5>} text="In order to view the Financial Assistance Broker Submissions, please follow the link below to request access" buttonText="Request Access" url="https://community.max.gov/x/fJwuRQ" />;
-            }
-            else if (permissionHelper.checkDabsReader(this.props.session)) {
-                // Access to FABS and DABS
-                firstBlock = <LandingBlock type={this.props.type} icon={<h5>DATA Act Broker Submission</h5>} text="Submit quarterly and monthly financial data." buttonText="Enter" url="#/landing" />;
-                secondBlock = <LandingBlock type={this.props.type} icon={<h5>Financial Assistance Broker Submission</h5>} text="Submit bi-monthly Financial Assistance award data." buttonText='Enter' url='#/detachedLanding' />;
-            }
-            else {
-                // Only access to FABS
-                firstBlock = <LandingBlock type={this.props.type} icon={<h5>DATA Act Broker Submission</h5>} text="In order to view the DATA Act Broker Submissions, please follow the link below to request access" buttonText="Request Access" url="https://community.max.gov/x/fJwuRQ" />;
-                secondBlock = <LandingBlock type={this.props.type} icon={<h5>Financial Assistance Broker Submission</h5>} text="Submit bi-monthly Financial Assistance award data." buttonText='Enter' url='#/detachedLanding' />;
-            }
+            firstBlock = <LandingBlock type={this.props.type} icon={<h5>DATA Act Broker Submission</h5>} text="Submit quarterly and monthly financial data." buttonText="Enter" url="#/landing" />;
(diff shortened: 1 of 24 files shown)


src/js/components/uploadDetachedFiles/UploadDetachedFileValidation.jsx
```diff
@@ -258,11 +260,14 @@ export default class UploadDetachedFileValidation extends React.Component {
...
 
 		validationBox = <ValidateDataFileContainer type={type} data={this.state.jobResults}/>;
-		if(!this.state.headerErrors && this.state.validationFinished) {
+		if (!this.state.headerErrors && this.state.validationFinished) {
 			validationBox = <ValidateValuesFileContainer type={type} data={this.state.jobResults} setUploadItem={this.uploadFile.bind(this)} updateItem={this.uploadFile.bind(this)} published={this.state.published}/>;
-			validationButton = <button className='pull-right col-xs-3 us-da-button' onClick={this.submitFabs.bind(this)}>Publish</button>;
-			if(this.state.published){
-				validationButton = <button className='pull-right col-xs-3 us-da-disabled-button' disabled>File Already Published</button>;
+
+			if (PermissionsHelper.checkFabsPermissions(this.props.session)) {
```
alburde1: Do we not want to show the button at all for DABS users or do we want to show the button but not let them click it with different text?

src/js/containers/history/HistoryContainer.jsx
```diff
@@ -17,7 +17,7 @@ class HistoryContainer extends React.Component {
 
     render() {
        return(
-        	<HistoryPage submissionID={this.props.params.submissionID}></HistoryPage>
+        	<HistoryPage submissionID={this.props.params.submissionID} type={this.props.route.type}></HistoryPage>
```
alburde1: Why does this one have opening and closing tags instead of just adding a `/` at the end of the opening one like everywhere else?

src/js/containers/router/RouterRoutes.jsx
```diff
@@ -70,26 +70,21 @@ const checkUserPermissions = (nextState, replace) => {
...
-    }
-    else if (session.user.helpOnly || (nextState.routes.length > 1 && nextState.routes[1].type == 'dabs' && !PermissionsHelper.checkDabsReader(session))) {
+    else if (session.user.helpOnly) {
         // if no permissions or attempting to reach DABS with improper permissions, bounce to help
         replace('/help');
     }
 }
 
-const checkReadUserPermissions = (nextState, replace) => {
+const checkDabsUserPermissions = (nextState, replace) => {
```
alburde1: If we're only using this in one place (checking if they can start a submission basically). Maybe this should be called `checkDabsUploadPermissions` because the name is misleading. I had to do a double-take when I was reading it because I was confused why we weren't allowing read-only users to get through the DABS permission check.

src/js/containers/router/RouterRoutes.jsx
```diff
@@ -317,7 +330,7 @@ const routeDefinitions = {
         },
         {
             path: 'detachedHelp',
-            onEnter: checkHelpUserPermissions,
+            onEnter: checkUserPermissions,
```
alburde1: It feels redundant to check permissions when entering the Help page. Really all we need to be doing is checking if they are logged in I think? Or do we not want them going to any of the DAIMS or practices or any of that if they have help only permissions? Because this check would always bounce them out of it to the DABS help page.

src/js/components/landing/BlockContent.jsx
```diff
@@ -22,21 +22,8 @@ export default class BlockContent extends React.Component {
         let thirdBlock = null;
 
         if (this.props.type === 'home') {
-            if (this.props.session.user.helpOnly) {
```
alburde1: Why are we removing the "no access" versions here? If they have helpOnly permissions they should still need to request access right?

michaeldhess: If they have helpOnly permissions, they should never get to the "home" page in the first place.

src/js/containers/router/RouterRoutes.jsx
michaeldhess: It seemed unnecessary to let users access the FABS help page if they don't have access to the FABS data in general, so it bumps them to the default Help. I wanted to keep the functionality as similar as possible to what it was before FABS was added to the Broker.

src/js/components/uploadDetachedFiles/UploadDetachedFileValidation.jsx
```diff
@@ -258,11 +260,16 @@ export default class UploadDetachedFileValidation extends React.Component {
...
+				validationButton = <button className='pull-right col-xs-3 us-da-button' onClick={this.submitFabs.bind(this)}>Publish</button>;
+				if (this.state.published){
+					validationButton = <button className='pull-right col-xs-3 us-da-disabled-button' disabled>File Already Published</button>;
```
alburde1: Follow-up question, do we want this version of the button to show whether or not the user has permissions and then, if it isn't published, check which version of the button to show? (publish vs no permissions)

PR #688, opened 2017-08-04
Db mk 2110 fabs download

kimminahm:
Downloadable files in fabs, also header warning files are also available to download for fabs AND dabs

### DIFF
--- a/src/js/components/validateData/ValidateDataFileComponent.jsx
+++ b/src/js/components/validateData/ValidateDataFileComponent.jsx
@@ -13,6 +13,8 @@ import * as ReviewHelper from '../../helpers/reviewHelper.js';
 import * as PermissionsHelper from '../../helpers/permissionsHelper.js';
 import * as GenerateFilesHelper from '../../helpers/generateFilesHelper.js';
 
+import UploadDetachedFilesError from '../uploadDetachedFiles/UploadDetachedFilesError.jsx';
+
 const propTypes = {
 
 };
@@ -31,7 +33,8 @@ export default class ValidateDataFileComponent extends React.Component {
             isError: false,
             hasFailed: false,
             signedUrl: '',
-            signInProgress: false
+            signInProgress: false,
+            error: null
         };
     }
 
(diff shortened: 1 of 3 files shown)


src/js/components/validateData/ValidateDataFileComponent.jsx
```diff
@@ -222,6 +225,72 @@ export default class ValidateDataFileComponent extends React.Component {
...
+                    this.openReport();
+                });
+            })
+            .catch((err) => {
+                this.setState({
+                    signInProgress: false
+                });
+                console.log(err);
+            });    
+        }else{
```
alburde1: Be consistent with your spacing. We usually put a space after the `else` so it looks like `else {` and also after if statements `if(truth check) {`. Also, stay consistent with where the else goes. MOST of the time, we do:
```
}
else {
```

src/js/components/validateData/ValidateDataFileComponent.jsx
```diff
@@ -222,6 +225,72 @@ export default class ValidateDataFileComponent extends React.Component {
...
+            console.log('Invalid File type selected: '+item.file_type)
```
alburde1: Just console logging if we don't have the right type isn't going to help anyone except a dev. There should be some indication to the user that something is wrong rather than the button just not working for some reason

src/js/components/validateData/ValidateDataFileComponent.jsx
```diff
@@ -222,6 +225,72 @@ export default class ValidateDataFileComponent extends React.Component {
...
+        }
+        
+    }
+
+    openReport() {
+        window.open(this.state.signedUrl);
+    }
+
+    clickedReport(item) {
+        console.log(item)
```
alburde1: Extraneous console log.

src/js/components/validateData/ValidateDataFileComponent.jsx
kimminahm: will work on it

src/js/components/validateData/ValidateDataFileComponent.jsx
```diff
@@ -312,7 +397,7 @@ export default class ValidateDataFileComponent extends React.Component {
                                     </div>
                                 </div>
                                 {uploadProgress}
-                                <div className="usa-da-validate-item-file-name">
+                                <div className='file-download' onClick={this.clickedReport.bind(this, this.props.item)} download={fileName} rel="noopener noreferrer">
```
alburde1: Font is a different size for file download links when there is a server/header error vs when there are just errors/warnings in the validations

PR #696, opened 2017-08-10
Db 2151 test environment flag

alburde1:
Adding a prod flag and setting it to false everywhere except the prod environment. Showing main landing page and test environment banner on all non-prod environments

- [ ] Reviewed by frontend
- [x] Configs updated to include the new global constant

Technical Release Notes:
- New global constant


dpb-bah:
Tested locally and confirmed AC

PR #727, opened 2017-09-19
Db mk 2194 help pages

kimminahm:
Adding the new help pages


src/_scss/pages/help/_content.scss
```diff
@@ -27,8 +27,16 @@
 	}
 
 	ul {
-		margin-top: 1.4rem;
-	}
+        margin-top: 1.5rem;
+        margin-bottom: 3.5rem;
+        margin-left: 0px;
+        ul {
```
alburde1: Why is there a `ul` inside a `ul`?

src/help/fabs_validations.csv
```diff
@@ -0,0 +1,28 @@
+Relates to Files,Rule Name,Rule Detail,Severity,Status,Type
+D2,D1,"Element value must be a valid enumeration as described in the ""Domain_Values.xlsx"" spreadsheet.",n/a,Implemented in ASP,Requirement
```
alburde1: We are aware that all of these aren't accurately named, labeled, or written correct?

src/js/components/help/daimsMessage.jsx
```diff
@@ -12,18 +12,18 @@ export default class DaimsMessage extends React.Component {
...
 
         return (
             <div className='daims'>
-                <span className='daims-header'>June 30, 2017: </span>
+                <span className='daims-header'>September 27, 2017: </span>
                 <span className='daims-message'>
                     Treasury released the DATA Act Information Model Schema (DAIMS) {version}.
-                    DAIMS {version} is a minor update of the schema and will be implemented in production in the fall of 2017.
-                    Find out more information about <a className='daims-link' href='https://community.max.gov/x/YgyuT'>DAIMS {version}</a>{type}
+                    DAIMS {version} is a major update of the schema and will be finalized in December 2017 for implementation in Q3 FY 2018.
```
alburde1: Finalized in Dec 2017 for use in FY of 2018? That doesn't sound right, are we sure that's right?

src/_scss/pages/help/_content.scss
kimminahm: Sub bullet points

src/help/fabs_validations.csv
kimminahm: These were copy pasted/taken from the existing rules. Will double check but I think this is it

src/js/components/help/daimsMessage.jsx
kimminahm: Thats the language in the docs I was provided. Will attach to Jira ticket

src/help/fabs_validations.csv
kimminahm: Words match up as far as I can tell with the provided table
https://github.com/fedspendingtransparency/data-act-broker-web-app/blob/dev/src/help/validations.csv

alburde1: Just making sure because things like `Must be blank for aggregate records; required for non-aggregate records.` have no meaning whatsoever and most if not all of these don't line up to the current rules

PR #729, opened 2017-09-19
Db mk 2194 help pages

kimminahm:


src/help/validations.csv
```diff
@@ -1,86 +1,62 @@
...
-A,A9,"ContractAuthorityAmountTotal_CPE= CPE aggregate value for GTAS SF 133 line #1540 + #1640, for the same reporting period",Fatal Error,Implemented,Calculation
-A,A10,BorrowingAuthorityAmountTotal_CPE= aggregate value for GTAS SF 133 line #1340 + #1440 for the same reporting period.,Fatal Error,Implemented,Calculation
-A,A11,SpendingAuthorityfromOffsettingCollectionsAmountTotal_CPE= aggregate value for GTAS SF 133 line #1750 + #1850 for the same reporting period.,Fatal Error,Implemented,Calculation
-A,A12,AdjustmentsToUnobligatedBalanceBroughtForward_CPE= aggregate value for GTAS SF 133 line #1010 through 1042 for the same reporting period.,Fatal Error,Implemented,Calculation
-A,A14,GrossOutlayAmountByTAS_CPE= value for GTAS SF 133 line #3020 for the same reporting period.,Fatal Error,Implemented,Calculation
-A,A15,UnobligatedBalance_CPE= value for GTAS SF 133 line #2490 for the same reporting period.,Fatal Error,Implemented,Calculation
-"A, B",A16,"All the elements that have FYB in file A & B would be expected in quarter 1, as a soft edit. It would be optional in subsequent periods.",Warning,Implemented,Requirement
+Relates to Files,Rule Name,Rule Detail,Severity,Type,Status
```
alburde1: Why did we put Status at the end and remove all the Type columns without removing the column header? This is going to break things

src/help/fabs_validations.csv
```diff
@@ -0,0 +1,101 @@
+Relates to Files,Rule Name,Rule Detail,Severity,Type,Status
```
alburde1: You have one extra header here too, Type should be gone

PR #743, opened 2017-09-27
Db mk 2246 publish duplicate

kimminahm:
- [ ] Dependent on backend https://github.com/fedspendingtransparency/data-act-broker-backend/pull/947
- [x] Reviewed by frontend


src/js/components/uploadDetachedFiles/UploadDetachedFileValidation.jsx
```diff
@@ -341,6 +343,10 @@ class UploadDetachedFileValidation extends React.Component {
 				validationButton = <button className='pull-right col-xs-3 us-da-disabled-button' disabled>You do not have permissions to publish</button>;
 			}
 		}
+
+		if (this.state.error !== 0) {
```
alburde1: Why was this moved down here instead of being left in the else if block?

src/js/components/uploadDetachedFiles/UploadDetachedFileValidation.jsx
```diff
@@ -102,14 +103,11 @@ class UploadDetachedFileValidation extends React.Component {
...
-				var submission = true;
-				if(response.publish_status==='published'){
-					submission = false;
-				}
+
 				const job = Object.assign({}, this.state.jobResults);
 				job.detached_award = response.jobs[0];
 
-				if (this.state.published && response.fabs_meta.publish_date && this.dataTimer) {
+				if (this.state.published !== 'publishing' && response.fabs_meta.publish_date && this.dataTimer) {
```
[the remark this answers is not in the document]
kimminahm: if it is unpublished it shouldnt keep checking, it should kill the loop so this works as expected

src/js/components/uploadDetachedFiles/UploadDetachedFileValidation.jsx
kimminahm: the logic above it is for if it is not publishing. It would change states before so this was the simplest solution I saw

PR #847, opened 2018-03-23
Fixing D file generation text

nmonga91:
* Updating stale text for ASP
* Removing unused practice and procedures help page & links since Broker links to MAX for that document


nmonga91:
@alburde1 I'll remove the CSS if any is associated with this page specifically

nmonga91:
There was no CSS file specifically for these pages so nothing was deleted.

alburde1:
There isn't any specific file, but in `_content.scss` there's a section for `.practices-procedures` and, looking at the search, it doesn't seem like this exists anywhere else. I'd look for other situations like that as well. This is one of those times when I think most of the css doesn't have its own dedicated file but is more scattered (or at least is in `_content.scss`

nmonga91:
@alburde1 removed files were checked for CSS that should've been removed, and CSS has been removed if applicable.

PR #856, opened 2018-04-16
Dev 648 rename fabs file types

alburde1:
- Replacing anything to do with D2_detached and detached_award with the clearer/more reasonable “fabs”
- Also removed fileTemplateName because we apparently aren’t using it anymore (if we were ever using it)

- [x] Reviewed by frontend
- [x] Merged concurrently with https://github.com/fedspendingtransparency/data-act-broker-backend/pull/1191


alburde1:
For the props/state, yes, I need to go do that right now. For the files, I didn't update the API calls, should the files be updated and not the calls, should it be both?

PR #890, opened 2018-06-20
Dev 1081 check status updates

alburde1:
Updating frontend to use new check_status endpoint.

- [x] Reviewed by frontend
- [ ] Merged concurrently with https://github.com/fedspendingtransparency/data-act-broker-backend/pull/1271

Technical Release Notes:
- Bugfix: Upload Corrected Files on cross-file page no longer active unless there is a staged file.


alburde1:
I'm not sure we have a solid convention. Get and Fetch are pretty much interchangeable I believe, Set should be if we're actually setting something specifically in that function and not doing anything else (at least that's how it felt to me) but I don't think we have any hard and fast rules about any of it.

PR #939, opened 2018-10-01
Filter dropdown

kojodoesdesign:
**High level description:**
Adds a "Created By" dropdown filter in the Broker Dashboard that allows the user to filter the table. 

**Technical details:**
- `CreatedByContainer` makes a call to the `list_all_submissions` endpoint, removes duplicate names, and passes the remaining values to populate the dropdown of an [Awesomeplete](https://www.npmjs.com/package/awesomplete) Typeahead component.
- `DropdownTypeahead` gets the value typed into the [Awesomeplete](https://www.npmjs.com/package/awesomplete) Typeahead component, and created an unchecked checkbox. 
- Upon Clicking the Checkbox, the staged `CreatedBy` filter tags are processed for display
- Processes applied `CreatedBy` filters and adds them to the `list_submissions` `filters` request param

**Link to JIRA Ticket:**
[DEV-1366](https://federal-spending-transparency.atlassian.net/browse/DEV-1366)

The following are ALL required for the PR to be merged:
- [x] Frontend review completed
- [x] Design review completed

PR #940, opened 2018-10-01
Date Modified dropdown

kojodoesdesign:
**High level description:**
Adds the "Last Date Modified" component to the Dashboard

**Technical details:**
- Add Date range component with Calendar
- Temporarily may reuse code from https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/939

**Link to JIRA Ticket:**
[DEV-1363](https://federal-spending-transparency.atlassian.net/browse/DEV-1363)


src/_scss/components/datepicker/_calendar-range-datepicker.scss
```diff
@@ -0,0 +1,261 @@
...
+    background-repeat: no-repeat;
+    color: #8B9898;
+    cursor: pointer;
+  }
+
+  .DayPicker-NavButton:hover {
+    opacity: 0.8;
+  }
+
+  // .DayPicker-NavButton--prev {
```
ebdabbs: If we don't need this code can you delete it?

src/_scss/pages/dashboard/_typeaheadFilter.scss
```diff
@@ -39,4 +39,57 @@
             }
         }
     }
-}
\ No newline at end of file
+
+  .filterdropdown {
+    .btn-default {
+      color: $color-white;
+      background-color: transparent;
```
ebdabbs: Can you add a fixed width so that the width of the button doesn't change when clicked (the font becomes bold)? (same for the Created By button)

src/js/components/SharedComponents/CalendarRangeDatePicker.jsx
```diff
@@ -0,0 +1,144 @@
...
+  const styleLeft = {
+    position: 'relative'
+  };
+  const styleRight = {
+    position: 'relative',
+    marginLeft: 'auto'
+  };
+  return (
+      <div className={className}>
+          <button style={styleLeft} onClick={() => onPreviousClick()}>
```
ebdabbs: Could you put each of the properties on a new line? Also, is there a reason `onClick={onPreviousClick}` doesn't work?

src/js/components/SharedComponents/CalendarRangeDatePicker.jsx
```diff
@@ -0,0 +1,144 @@
...
+import DayPicker, { DateUtils } from 'react-day-picker';
+
+import Moment from 'moment';
+
+function Navbar({
+  onPreviousClick,
+  onNextClick,
+  className
+}) {
+  const styleLeft = {
```
ebdabbs: Inline styles should be avoided where possible. Could you use a modifier class and add these properties with CSS?

src/js/components/SharedComponents/CalendarRangeDatePicker.jsx
```diff
@@ -0,0 +1,144 @@
+import React, { PropTypes } from 'react';
+import DayPicker, { DateUtils } from 'react-day-picker';
+
+import Moment from 'moment';
+
+function Navbar({
```
ebdabbs: It looks like there are two components in this file, `NavBar` and `CalendarRangeDatePicker`. Please move `NavBar` to it's own file. (See the components section at https://github.com/fedspendingtransparency/data-act-documentation/blob/master/frontend_apps/SoD.md#overview)

src/js/redux/actions/lastDateModifiedActions.js
```diff
@@ -0,0 +1,4 @@
+export const setLastDateModifiedList = (state) => ({
+  type: 'SET_LASTDATEMODIFIED_LIST',
```
ebdabbs: I would recommend using underscores between each word to be consistent with other action names.

src/js/redux/reducers/dashboard/dashboardFilterReducer.js
```diff
@@ -6,113 +6,123 @@
...
+        [action.dashboard]: dashboard
+      });
+    }
+    case 'TOGGLE_DASHBOARD_FILTER': {
+      // get the existing list
+      let list;
+      if (!Array.isArray(state[action.dashboard][action.table][action.filter])) {
+        const Objectlist = action.value;
+        const prevFilter = state[action.dashboard][action.table][action.filter];
+        if (prevFilter.startDate === Objectlist.startDate && prevFilter.endDate === Objectlist.endDate) {
```
ebdabbs: `TOGGLE_DASHBOARD_FILTER` is designed to update filters that are stored as arrays of strings.  Instead of modifying it to fit the needs of the date modified filter, you could use `UPDATE_DASHBOARD_FILTER` to store the object with a start date and end date, or create a new action.

src/js/components/SharedComponents/CalendarRangeDatePicker.jsx
```diff
@@ -0,0 +1,109 @@
...
+      if (this.props.minmaxDates.maxDate && this.props.minmaxDates.minDate) {
+          return (<DayPicker
+              showOutsideDays
+              fixedWeeks
+              fromMonth={this.props.minmaxDates.minDate}
+              toMonth={this.props.minmaxDates.maxDate}
+              navbarElement={<Navbar />}
+              className="innerCalendarDatePicker"
+              numberOfMonths={this.props.numberOfMonths}
+              selectedDays={[from, { from, to }]}
```
ebdabbs: `DayPicker` is expecting `selectedDays` to be a function rather than an array. See use of `DayPicker` [here](https://github.com/fedspendingtransparency/data-act-broker-web-app/blob/dev/src/js/components/generateFiles/components/DatePicker.jsx#L238).

kojodoesdesign: The `DayPicker` library documentation shows that `selectedDays` can also be an array. 
Example: http://react-day-picker.js.org/examples/selected/

src/js/containers/dashboard/LastDateModifiedContainer.jsx
```diff
@@ -0,0 +1,97 @@
...
+    this.loadData();
+  }
+
+  loadData() {
+    if (_.isEmpty(this.props.lastDateModifiedList.lastDateModified)) {
+      // we need to populate the list
+      if (this.props.detached) {
+        lastDateModifiedHelper.fetchLastDateModified()
+          .then((data) => {
+            this.props.setLastDateModifiedList(data);
```
ebdabbs: If all you need is the min and max modified date, it would be a better pattern to pass `data` to a parse function and store only what you need in the component state, rather than the entire API response in Redux.

src/js/components/SharedComponents/CalendarRangeDatePicker.jsx
```diff
@@ -0,0 +1,109 @@
...
+      maxDate: '',
+      minDate: ''
+  }
+};
+
+export default class CalendarRangeDatePicker extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.dateValues = {};
```
ebdabbs: `this.dateValues` is redundant because it gets set to `this.state` on line 66. When it is referenced in `sendToFilters`, you can lookup `this.state.to` / `this.state.from` instead.

src/js/components/SharedComponents/CalendarRangeDatePicker.jsx
ebdabbs: Resolved by updating react-day-picker from v 2.2.3 to v 7.2.4

PR #1027, opened 2019-03-21
Update no resultsmessage loadingmessage

jonhill13:
**High level description:**
Updates Broker submission table's loading message and and no results message

**Technical details:**
> • pull in USASpending loading message and no results message
> • update CSS
> • update table logic

**Link to JIRA Ticket:**
[DEV-1283](https://federal-spending-transparency.atlassian.net/browse/DEV-1283)

- [x] Design Review Completed ( pictures are in the design ticket [ticket](https://federal-spending-transparency.atlassian.net/browse/DEV-2300) )


src/js/components/dashboard/DashboardTable.jsx
```diff
@@ -400,7 +399,8 @@ export default class DashboardTable extends React.Component {
                         onSort={this.sortTable.bind(this)} />
                 </div>
                 <div className="text-center">
-                    {this.state.message}
+                    {this.state.noResults && <NoResultsMessage />}
+                    {this.props.isLoading && <LoadingMessage />}
                 </div>
```
MaxwellKendall: 👍 I like this approach better. I guess formerly we were doing a bunch of CSS magic based on the class value in `line 393`?

src/js/components/SharedComponents/LoadingMessage.jsx
```diff
@@ -0,0 +1,63 @@
...
+//                                 {this.props.loadingMessage}
+//                             </div>
+//                         </div>
+//                     </div>
+//                 </CSSTransitionGroup>
+//             </div>
+//         );
+//     }
+// };
+
```
MaxwellKendall: Probably delete this unless you wanted to keep it here for some reason.

src/_scss/abstract/variables/images.scss
```diff
@@ -0,0 +1,2 @@
+// image path
+$image-path: '/img' !default;
```
MaxwellKendall: This is a cool SASS Feature i didn't know about.

src/js/components/dashboard/DashboardTable.jsx
```diff
@@ -204,16 +208,16 @@ export default class DashboardTable extends React.Component {
 
         const headerClasses = classes;
 
-        let message = '';
+        let noResults = '';
         if (this.props.data.length === 0) {
```
MaxwellKendall: since `noResults` is a boolean, I suggest making it's initial value `false`

src/js/components/dashboard/DashboardTable.jsx
jonhill13: we were adding a class to the table header based on loading or not loading or results. I originally deleted that, however, that was taking care of the opacity in the table header which is needed to make the design.  I added it back in with `classnames` npm package and `tableClasses` method

src/js/components/dashboard/DashboardTable.jsx
```diff
@@ -389,7 +394,7 @@ export default class DashboardTable extends React.Component {
 
         return (
             <div className="usa-da-submission-list">
-                <div>
+                <div className={this.tableClasses()}>
                     <FormattedTable
```
MaxwellKendall: So you can do this:

```javascript

classNames({
'name-of-class':boolean-producing-javascript-expression
})
```

MaxwellKendall: Also, I don't think you need to put this logic in a class method, you can just do

`<div className={`some-class-thats-always-here ${classNames({ not-always-here: props.condition })}`}>`

jonhill13: Yeah, [classnames](https://www.npmjs.com/package/classnames) provides a clean way to handle multiple classnames with logical operations

`var classNames = require('classnames');
 
var Button = React.createClass({
  // ...
  render () {
    var btnClass = classNames({
      btn: true,
      'btn-pressed': this.state.isPressed,
      'btn-over': !this.state.isPressed && this.state.isHovered
    });
    return <button className={btnClass}>{this.props.label}</button>;
  }
});`

MaxwellKendall:
@ebdabbs want to make sure to get your 👀 on these two new npm modules coming through with this PR.

Links to the new lines in the new `package.json` below:

- [class-names](https://github.com/fedspendingtransparency/data-act-broker-web-app/blob/21106aed7c6b0c38e8bf72502e2cf3967c54a8eb/package.json#L50)
- [React-Transition-Groups](https://github.com/fedspendingtransparency/data-act-broker-web-app/blob/21106aed7c6b0c38e8bf72502e2cf3967c54a8eb/package.json#L133)

src/js/components/dashboard/DashboardTable.jsx
MaxwellKendall: Perfect, lets just double check these new modules are good to go, then I'll approve.

PR #1029, opened 2019-03-21
DEV-1308: detached file url generation

alburde1:
**High level description:**
Forgot to account for detached files in the file generation on click

**Technical details:**
Had to add a new endpoint to account for how both detached file D and detached file A generation work.

**Link to JIRA Ticket:**
[DEV-1308](https://federal-spending-transparency.atlassian.net/browse/DEV-1308)

- [x] Merged after [Backend#1548](https://github.com/fedspendingtransparency/data-act-broker-backend/pull/1548)

### DIFF
--- a/src/js/components/generateDetachedFiles/GenerateDetachedFilesPage.jsx
+++ b/src/js/components/generateDetachedFiles/GenerateDetachedFilesPage.jsx
@@ -48,12 +48,10 @@ export default class GenerateDetachedFilesPage extends React.Component {
                     header: '',
                     description: ''
                 },
-                download: {
-                    show: false,
-                    url: ''
-                },
+                showDownload: false,
                 valid: false,
-                status: ""
+                status: "",
+                jobID: null
             },
             d2: {
                 startDate: null,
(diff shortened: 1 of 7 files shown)


src/js/components/generateDetachedFiles/GenerateDetachedFilesPage.jsx
```diff
@@ -308,7 +313,8 @@ export default class GenerateDetachedFilesPage extends React.Component {
                 handleDateChange={this.handleDateChange.bind(this)}
                 updateError={this.updateError.bind(this)}
                 generateFile={this.generateFile.bind(this)}
-                toggleAgencyType={this.toggleAgencyType} />);
+                toggleAgencyType={this.toggleAgencyType}
+                clickedDownload={this.clickedDownload.bind(this)} />);
```
jonhill13: My main concern is binding in the prop. `This is bad for performance, as it may cause unnecessary re-renders if a brand new function is passed as a prop to a component that uses reference equality check on the prop to determine if it should update.` But after looking throughout the application it looks like we are binding in the prop a lot of places.

PR #1038, opened 2019-03-29
DEV-2329 Update D1/D2 funding toggle within DABS

jonhill13:
**High level description:**
Add funding toggle to D1 and D2 file types.

**Technical details:**
`fundingAgency` state property is now part of `d1` and `d2` state properties.

**Link to JIRA Ticket:**
[DEV-2329](https://federal-spending-transparency.atlassian.net/browse/DEV-2329)

### DIFF
--- a/src/js/components/generateDetachedFiles/GenerateDetachedFilesPage.jsx
+++ b/src/js/components/generateDetachedFiles/GenerateDetachedFilesPage.jsx
@@ -7,6 +7,7 @@ import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
 
 import React, { PropTypes } from 'react';
 import moment from 'moment';
+import { assign } from 'lodash';
 import Navbar from '../SharedComponents/navigation/NavigationComponent';
 import Footer from '../SharedComponents/FooterComponent';
 import AgencyListContainer from '../../containers/SharedContainers/AgencyListContainer';
@@ -39,7 +40,6 @@ export default class GenerateDetachedFilesPage extends React.Component {
             showDateSelect: false,
             showSubmitButton: false,
             buttonDisabled: true,
-            fundingAgency: false,
             d1: {
                 startDate: null,
                 endDate: null,
@@ -51,7 +51,8 @@ export default class GenerateDetachedFilesPage extends React.Component {
                 showDownload: false,
                 valid: false,
                 status: "",
-                jobID: null
+                jobID: null,
+                fundingAgency: false
             },
             d2: {
                 startDate: null,
@@ -64,7 +65,8 @@ export default class GenerateDetachedFilesPage extends React.Component {
                 showDownload: false,
                 valid: false,
                 status: "",
-                jobID: null
+                jobID: null,
+                fundingAgency: false
             }
         };
 
@@ -192,9 +194,11 @@ export default class GenerateDetachedFilesPage extends React.Component {
         });
     }
 
-    toggleAgencyType() {
+    toggleAgencyType(type) {
+        const newType = assign({}, this.state[type]);
+        newType.fundingAgency = !newType.fundingAgency;
         this.setState({
-            fundingAgency: !this.state.fundingAgency
+            [type]: newType
         });
     }
 
@@ -202,7 +206,7 @@ export default class GenerateDetachedFilesPage extends React.Component {
         // generate specified file
         const cgacCode = this.state.codeType !== 'frec_code' ? this.state.agency : '';
         const frecCode = this.state.codeType === 'frec_code' ? this.state.agency : '';
-        const agencyType = this.state.fundingAgency ? 'funding' : 'awarding';
+        const agencyType = this.state[file].fundingAgency ? 'funding' : 'awarding';
 
         const tmpFile = Object.assign({}, this.state[file]);
         tmpFile.status = "generating";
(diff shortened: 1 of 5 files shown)


src/js/containers/generateFiles/GenerateFilesContainer.jsx
```diff
@@ -338,9 +339,11 @@ class GenerateFilesContainer extends React.Component {
...
     }
 
-    toggleAgencyType() {
+    toggleAgencyType(type) {
+        const newType = _.assign({}, this.state[type]);
+        newType.fundingAgency = !newType.fundingAgency;
         this.setState({
-            fundingAgency: !this.state.fundingAgency
+            [type]: newType
         });
```
MaxwellKendall: Some feedback on implementation:

`setState` comes with the `prevState` parameter you can use to accomplish the same thing without having to create a new const:

```javascript
this.setState((prevState) => ({
  [type]: {
     ...prevState[type],
    isFundingAgency: !prevState[type].isFundingAgency
  }
})
```

src/js/containers/generateFiles/GenerateFilesContainer.jsx
```diff
@@ -56,7 +56,8 @@ class GenerateFilesContainer extends React.Component {
                     header: '',
                     description: ''
                 },
-                showDownload: false
+                showDownload: false,
+                fundingAgency: false
             },
```
MaxwellKendall: Recommend changing state property `d1/d2.fundingAgency` to `d1/d2.isFundingAgency` b/c it's a boolean and prepending `is` makes that a bit more readable IMO.

src/js/components/generateFiles/GenerateFilesContent.jsx
```diff
@@ -95,50 +122,73 @@ export default class GenerateFilesContent extends React.Component {
...
                             startingTab={1}
-                            value={this.props.d1}
-                            error={this.props.d1.error}
-                            showDownload={this.props.d1.showDownload}
+                            value={d1}
+                            error={d1.error}
+                            showDownload={d1.showDownload}
                             onDateChange={this.handleDateChange.bind(this, "d1")}
                             updateError={this.updateError.bind(this, "d1")}
                             clickedDownload={this.clickedDownload.bind(this, "D1")} />
```
MaxwellKendall: General tech debt here -- putting `.bind()` in the render creates a new anonymous fn each render. This is everywhere in broker and IMO should be put in the constructor for performance reasons:

https://reactjs.org/docs/faq-functions.html

src/js/components/generateFiles/GenerateFilesContent.jsx
```diff
@@ -95,50 +122,73 @@ export default class GenerateFilesContent extends React.Component {
...
+                                        onFocus={this.showTooltip.bind(this, 'd1')}
+                                        onBlur={this.closeTooltip.bind(this, 'd1')}
+                                        onMouseLeave={this.closeTooltip.bind(this, 'd1')}
+                                        onMouseEnter={this.showTooltip.bind(this, 'd1')} >
+                                        <InfoCircle />
+                                    </button>
+                                </div>
+                            </span>
+                        </div>
                         <GenerateFileBox
```
MaxwellKendall: A lot of duplicate code here, perhaps could be abstracted into class method.

src/js/components/generateFiles/GenerateFilesContent.jsx
```diff
@@ -55,34 +61,55 @@ export default class GenerateFilesContent extends React.Component {
...
     }
 
-    closeTooltip() {
+    closeTooltip(type) {
+        const toolTipState = assign({}, this.state.showInfoTooltip);
+        toolTipState[type] = !toolTipState[type];
         this.setState({
-            showInfoTooltip: false
+            showInfoTooltip: toolTipState
         });
```
MaxwellKendall: Could also use prevState here.

src/js/components/generateDetachedFiles/GenerateDetachedFilesPage.jsx
```diff
@@ -192,17 +194,19 @@ export default class GenerateDetachedFilesPage extends React.Component {
...
     }
 
-    toggleAgencyType() {
+    toggleAgencyType(type) {
+        const newType = assign({}, this.state[type]);
+        newType.fundingAgency = !newType.fundingAgency;
         this.setState({
-            fundingAgency: !this.state.fundingAgency
+            [type]: newType
         });
```
MaxwellKendall: Could also use `prevState` here.

src/js/containers/generateFiles/GenerateFilesContainer.jsx
jonhill13: Thank you.  I was told by @ebdabbs not to update properties in `setState`.

src/js/containers/generateFiles/GenerateFilesContainer.jsx
jonhill13: Agreed

src/js/components/generateFiles/GenerateFilesContent.jsx
jonhill13: Thank you. I am aware binding in the render is a horrible practice. Also, I am a believer of not piling crap on top of crap.  I am also not a fan of binding in the constructor or binding when it is not necessary at all. However, my first attempt at this PR included pulling in more babel presets allowing us to write arrow functions in our classes keeping `this` which would fix both binding in render and binding at all, was biting off more than I could chew at this moment due to the fact that we are continuously wrapping props in other methods and then binding again further down the tree. Due to the time constraint of this sprint, I took a step back and instead of refactoring multiple components, I decided to complete the task of this ticket. I do believe we need to have a discussion of refactoring not only the binds but a lot of this web app.

src/js/components/generateFiles/GenerateFilesContent.jsx
jonhill13: Yeah, let me take another look at this.

PR #1067, opened 2019-06-11
[DEV-2536] Adding Unit tests for helper methods that are (1) not REST requests & (2) exported

MaxwellKendall:
**High level description:**

Adds unit tests for every exported helper function. Added a test file for each helper file for read-ability and to acknowledge possible additional tests.

**Technical details:**
Trying to test the scenario for each distinct return value.

**Link to JIRA Ticket:**
[DEV-2536](https://federal-spending-transparency.atlassian.net/browse/DEV-2536)

- [x] Unit tests added (if applicable)
- [x] Passes linter

### DIFF
--- a/tests/helpers/generateFilesHelper-test.js
+++ b/tests/helpers/generateFilesHelper-test.js
@@ -1,0 +1,6 @@
+// No tests needed ATM; all API Requests
+describe('nothing to test ATM', () => {
+    it('doesn\'t require testing ATM', () => {
+        expect(true).toEqual(true);
+    });
+});
--- a/tests/helpers/helpHelper-test.js
+++ b/tests/helpers/helpHelper-test.js
@@ -1,0 +1,7 @@
+// No tests needed ATM; all API Requests
+// parseMarkdown, unescapeInlineHtml?
+describe('nothing to test ATM', () => {
+    it('doesn\'t require testing ATM', () => {
+        expect(true).toEqual(true);
+    });
+});
--- a/tests/helpers/reviewHelper-test.js
+++ b/tests/helpers/reviewHelper-test.js
@@ -1,29 +1,75 @@
-import * as reviewHelper from 'helpers/reviewHelper';
+import * as reviewHelper from "helpers/reviewHelper";
 
-describe('reviewHelper', () => {
-    describe('getFileStates', () => {
+describe("reviewHelper", () => {
+    describe("getFileStates", () => {
         const generateParam = (jobs = []) => ({
             jobs
         });
         const defaultResult = (name) => ({
-			report: null,
-			error_count: 0,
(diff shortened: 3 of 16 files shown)


tests/helpers/generateFilesHelper-test.js
```diff
@@ -0,0 +1,6 @@
+// No tests needed ATM; all API Requests
```
ebdabbs: `generateFile` has some logic to create the `callBody` that could be tested, but ultimately we should move that logic out of the helper function.

tests/helpers/helpHelper-test.js
```diff
@@ -0,0 +1,7 @@
+// No tests needed ATM; all API Requests
+// parseMarkdown, unescapeInlineHtml?
```
ebdabbs: I think these two functions should be tested - they are different from most other helpers and might require a different format.

tests/helpers/reviewHelper-test.js
```diff
@@ -0,0 +1,75 @@
+import * as reviewHelper from "helpers/reviewHelper";
+
+describe("reviewHelper", () => {
+    describe("getFileStates", () => {
```
ebdabbs: Could we also test `determineExpectedPairs`?

tests/helpers/submissionListHelper-test.js
```diff
@@ -0,0 +1,7 @@
+// No tests needed ATM; all API Requests
+// parseRecentActivity?
```
ebdabbs: There was a bug that came up recently related to file size calculations (some of which happen in `parseRecentActivity`), so it should have unit tests. Or we move a generic file size calculation to it's own helper function.

tests/helpers/uploadHelper-test.js
```diff
@@ -0,0 +1,6 @@
+// No tests needed ATM; all API Requests
```
ebdabbs: `prepareMetadata` could be tested

tests/helpers/generateFilesHelper-test.js
MaxwellKendall: Would like to disregard this test at the moment per above comment and exclude from scope of this story.

tests/helpers/helpHelper-test.js
MaxwellKendall: @ebdabbs ,

I think mocking the data for these functions is a big effort. Not sure what's going on here and seems like we're using a third party library `markdown` to do some dirty work here.

These are the functions I mentioned that I tried to test but mocking the data for the parameters proved troublesome.

ebdabbs: https://federal-spending-transparency.atlassian.net/browse/DEV-2918

PR #1071, opened 2019-06-18
Sprint 84: Dev -> Staging

alburde1:


dpb-bah:
* Migrated front-end architecture to use Webpack4.
* Updated front-end focused elements to not click when any key is pressed.

PR #1081, opened 2019-07-09
[dev-3020] Configuring Google Analytics for broker w/ new tracking id

MaxwellKendall:
**High level description:**
Updating `npm` Goolge Analytics package to latest version which has no breaking changes. [docs](https://www.npmjs.com/package/react-ga)

**Technical details:**
Adds conditions to init GA in `RouterContainer` based on `NODE_ENV` and if the tracking ID exists. Also have branch for config repo to updated the tracking ID [here](https://github.com/fedspendingtransparency/data-act-broker-config/compare/new_ga_tracking_id).

If we were to merge this branch into production today (the config-repo branch), it would begin logging w/ the new tracking ID in production only.

**Link to JIRA Ticket:**
[DEV-3020](https://federal-spending-transparency.atlassian.net/browse/DEV-3020)

- [x] #1077 merged


MaxwellKendall:
> This is a good start to collecting basic info. Pageviews will require additional logic (since it functions as a SPA) - see implementation in the [usaspending router](https://github.com/fedspendingtransparency/usaspending-website/blob/dev/src/js/containers/router/RouterContainer.jsx#L53). This will also log every environment (local, dev, staging, and prod) so we'll want to determine how to best filter out non-prod environments, either in the codebase or with GA Segments.

@ebdabbs I recommend the following:

Merge [this branch](https://github.com/fedspendingtransparency/data-act-broker-config/compare/new_ga_tracking_id) from the config repo to master to begin tracking with the new ID in production.

If we do this, the lower environments will no longer have `ga` initialized w/ a tracking ID. Instead, it will be invoked w/out a tracking id which results in a console.warning, but that is all. We can also put this branch in dev/staging and `ga` will only be initialized when the tracking id is defined and the node_env is production.

PR #1098, opened 2019-08-13
Sprint 88: Dev -> Staging

alburde1:


dpb-bah:
* Updated architecture to use a single global constants file for simplicity and clarity.
* Updated dialogue when requesting to delete a submission on the dashboard.
* Updated Last Modified Filter for the submission dashboard to exclude dates before the earliest modified date.

PR #1135, opened 2019-09-23
Sprint 91: Dev -> Staging

alburde1:


dpb-bah:
* Added link to Download All Submission Comments file on Review and Certify Data page.
* Updated logic to properly list the file size of user uploads.
* Reworked tests and cleaned up code removing deprecated lifecycle methods.

PR #1151, opened 2019-10-11
DEV-3598 Retain DABS submission step on refresh

ebdabbs:
**High level description:**
Fixes a bug that caused the step to change when refreshing during the DABS submission process. 

**Technical details:**
- We were making a new API call every time the component mounts, which automatically redirected to the last completed step. The solution was to check whether a step is already indicated in the url params when `componentDidMount` is invoked
- Also refactored the state property `completedSteps` into an array of booleans instead of an object keyed by indices
- Removed redundant `routeParams` prop
- Updated unit tests to prevent mutating the mocked state and props, and to account for the changes above

**Link to JIRA Ticket:**
[DEV-3598](https://federal-spending-transparency.atlassian.net/browse/DEV-3598)

- [x] Scheduled Demo including Design/Testing/Front-end OR Provided Instructions for Local Testing above and in JIRA (if applicable)
- N/A All componentWillReceiveProps, componentWillMount, and componentWillUpdate in relevant child/parent components/containers replaced with [future compatible life-cycle methods](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html) per [JIRA item 3339](https://federal-spending-transparency.atlassian.net/browse/DEV-3339)
- [x] Link to this PR in JIRA ticket

### DIFF
--- a/src/js/containers/submission/SubmissionContainer.jsx
+++ b/src/js/containers/submission/SubmissionContainer.jsx
@@ -45,13 +45,7 @@ export class SubmissionContainer extends React.Component {
     }
 
     componentDidMount() {
-        if (this.props.params.type) {
-            // Refreshed the page, stay on the step indicated by the url
-            this.getSubmission(true);
-        }
-        else {
-            this.getSubmission();
-        }
+        this.getSubmission(this.props.params.type);
     }
 
     componentDidUpdate(prevProps) {
@@ -155,7 +149,7 @@ export class SubmissionContainer extends React.Component {
         let step = this.state.step;
         step += 1;
         const completedSteps = [];
-        for (let i = 0; i < 4; i++) {
+        for (let i = 0; i < this.state.completedSteps.length; i++) {
             completedSteps[i] = step >= i;
         }
         this.setState({ step, completedSteps }, this.updateRoute);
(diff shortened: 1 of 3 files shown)


src/js/containers/submission/SubmissionContainer.jsx
```diff
@@ -51,7 +45,13 @@ export class SubmissionContainer extends React.Component {
     }
 
     componentDidMount() {
-        this.getSubmission();
+        if (this.props.params.type) {
```
ebdabbs: simplify to pass this.props.params.type as the bool value

src/js/containers/submission/SubmissionContainer.jsx
```diff
@@ -143,10 +154,10 @@ export class SubmissionContainer extends React.Component {
     nextStep() {
         let step = this.state.step;
         step += 1;
-        const completedSteps = reduce(this.state.completedSteps, (acc, value, key) => {
-            if (step.toString() === key) acc[key] = true;
-            return acc;
-        }, this.state.completedSteps);
+        const completedSteps = [];
+        for (let i = 0; i < 4; i++) {
```
ebdabbs: Use length of this.state.completedSteps instead of hard-coded 4

PR #1156, opened 2019-10-18
DEV-3658 dashboard fiscal year filter

ebdabbs:
**High level description:**
Adds a fiscal year filter to the dashboard page. 

**Technical details:**
- Makes an API call to `/latest_certification_period` endpoint from both quarter filter and fiscal year filter containers to determine how many FY checkboxes to render and to support the following logic: 
  - Disable future quarters if the current fiscal year is the only FY selected
  - Disable Q1 button if 2017 is the only FY selected
  - Disable the current fiscal year checkbox if the only quarters selected are greater than the latest quarter that could have been certified for the current FY
   - Disable FY 17 if Q1 is the only quarter currently selected

**Link to JIRA Ticket:**
[DEV-3658](https://federal-spending-transparency.atlassian.net/browse/DEV-3658)

**Mockup**
https://bahdigital.invisionapp.com/share/7GIABMNX8MV#/296021489_Historical-Search-Monthly-Single-Data

- [x] Design review
- [x] Demo including Design/Testing/Front-end OR Provided Instructions for Local Testing above and in JIRA
- [x] Merged concurrently with [Backend #1703](https://github.com/fedspendingtransparency/data-act-broker-backend/pull/1703)
- N/A All componentWillReceiveProps, componentWillMount, and componentWillUpdate in relevant child/parent components/containers replaced with [future compatible life-cycle methods](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html) per [JIRA item 3339](https://federal-spending-transparency.atlassian.net/browse/DEV-3339)
- [x] Link to this PR in JIRA ticket

### DIFF
--- a/src/_scss/pages/dashboard/filters/_fiscalYears.scss
+++ b/src/_scss/pages/dashboard/filters/_fiscalYears.scss
@@ -1,8 +1,10 @@
 ul.fiscal-years {
     @include unstyled-list;
-    @include display(flex);
     font-weight: $font-normal;
     padding-top: rem(15);
+    .fiscal-years__wrapper {
+        @include display(flex);
+    }
     .fiscal-years__left {
         padding-right: rem(4);
     }
--- a/src/_scss/pages/dashboard/filters/_fiscalYears.scss
+++ b/src/_scss/pages/dashboard/filters/_fiscalYears.scss
@@ -1,0 +1,3 @@
+ul.fiscal-years {
+    @include unstyled-list;
+}
--- a/src/_scss/pages/dashboard/filters/_fiscalYears.scss
+++ b/src/_scss/pages/dashboard/filters/_fiscalYears.scss
@@ -1,3 +1,27 @@
 ul.fiscal-years {
     @include unstyled-list;
+    @include display(flex);
+    font-weight: $font-normal;
+    padding-top: rem(15);
+    .fiscal-years__left {
+        padding-right: rem(4);
+    }
+    .fiscal-years__right {
+        padding-left: rem(4);
+    }
+    li.fy-option {
+        &:before {
+            display: none;
+        }
+        &:after {
+            display: none;
+        }
+        list-style-type: none;
+        .fy-option__label {
+            font-weight: $font-normal;
+            font-size: $small-font-size;
+            line-height: rem(26);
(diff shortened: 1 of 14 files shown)


ebdabbs:
Design & testing review feedback:
- Align horizontal dividers under page heading and "Filters" heading
- Change "Rule" to "Rules" for consistency with "Files"
- Ensure future quarters are disabled after current FY is selected

src/_scss/pages/dashboard/filters/_fiscalYears.scss
```diff
@@ -0,0 +1,29 @@
...
+    font-weight: $font-normal;
+    padding-top: rem(15);
+    .fiscal-years__wrapper {
+        @include display(flex);
+    }
+    .fiscal-years__left {
+        padding-right: rem(4);
+    }
+    .fiscal-years__right {
+        padding-left: rem(4);
```
alburde1: Is there a reason we can't just put padding on the right for each column and not have to specify which side something is on? Seems simpler for future when we have more than 2 columns (I assume we're increasing column count? Maybe we won't, it just seems like a lot of empty space on the right)

src/js/components/dashboard/filters/FiscalYearFilter.jsx
```diff
@@ -0,0 +1,96 @@
...
+                if (this.props.selectedFY.includes(fy.year)) {
+                    this.props.pickedFy(fy.year);
+                }
+            });
+        }
+    }
+
+    render() {
+        let allFY = true;
+        const leftCount = Math.ceil(this.props.allFy.length / 2);
```
alburde1: Just bringing up the question created by the earlier comment, is it always going to be 2 columns or will it become 3 at a certain point? If it's staying at 2 then my earlier comment is probably also extraneous.

ebdabbs: I think it will stay two columns since we're modeling it after advanced search: 
TimePeriod

alburde1: Maybe it's the larger gap between them that makes it look like less of an empty space on the side. That makes sense though. In that case ignore the previous comment as well.

PR #1181, opened 2019-11-15
Declaring Component Library as Dependency

MaxwellKendall:
**High level description:**
import components from the library w/ the styles included as follows:

`import { Component } from data-transparency-ui;`

**Technical details:**
- Adds config to webpack to parse JS/JSX files from node-modules in this dependency only.
- Includes CSS from new package
- Updates `-lock` file
- Changes dev-server to port `3002` for npm start script and auth-call back ENV Variable

- [x] Bundle Size audit


MaxwellKendall:
@ebdabbs,

Want to get this out there, but would like to make sure that, for instance, we're not using `lodash` twice. Might have to change some config to the CL itself to make sure.

MaxwellKendall:
@ebdabbs this should be good to go!

PR #1206, opened 2019-12-03
Fix graph legend

ebdabbs:
**High level description:**
Fixes a bug that caused the legend to sometimes be generated incorrectly.

**Technical details:**
The previous implementation expected the rule label to be the first property in each rule object. The fix references the property name instead of using an index.

**Link to JIRA Ticket:**

[DEV-3664](https://federal-spending-transparency.atlassian.net/browse/DEV-3664)

Reviewer(s):


dpb-bah:
Can confirm this is the only instance of this bug (taking the first value of an incoming object).

PR #1207, opened 2019-12-03
Sprint 96: Dev -> Staging

alburde1:


dpb-bah:
Added an Agency Data Dashboard page with the following features:
* View an agency’s certified submissions and warning data to discover trends over a certain period of time.
* Filter data by agency (based on the user’s permissions), reporting period, files, and DAIMS rules to pinpoint specific warnings.
* Track each warning to individual downloadable files certified by the user’s agencies.

PR #1234, opened 2019-12-17
Sprint 97 Dev -> Staging

alburde1:


dpb-bah:
* Added "Go-To-Page" functionality for the Agency Data Dashboard Table
* Updated Generate D1/D2 file pages to allow users to download pipe-delimited TXT files
* Resolved various minor bugs.

PR #1257, opened 2020-01-17
DEV-3340 remove componentWillReceiveProps

ebdabbs:
**High level description:**

Removes instances of the deprecated React lifecycle method `componentWillReceiveProps` and subsequent refactoring.

**Technical details:**

- For multiple instances, refactoring involved removing the anti-pattern of storing props in state when the props could be referenced directly. 
- This PR also moves the functions that load markdown for the help pages to the container, so they only need to be called once and the markdown can be passed down to the various pages via props

**Link to JIRA Ticket:**

[DEV-3340](https://federal-spending-transparency.atlassian.net/browse/DEV-3340)

**Testing**
- No impact to end user

Author: 
- [x] Linked to this PR in JIRA ticket
- [x] Provided Instructions for Local Testing above and in JIRA

Reviewer(s):


src/js/components/validateData/ValidateDataFileComponent.jsx
```diff
@@ -442,7 +434,7 @@ export default class ValidateDataFileComponent extends React.Component {
                                         tabIndex={0}
                                         className={clickDownloadClass}
                                         onKeyDown={clickDownloadOnKeyDownHandler}
-                                        onClick={clickDownload}
```
alburde1: Here we have `clickDownload` which may be `null` at the end of all of this, affected by `canDownload` but we didn't change `clickedReport` at all so how are we covering for this variable?

src/js/components/history/HistoryTable.jsx
```diff
@@ -6,27 +6,20 @@
...
-const defaultProps = {
-    submissionID: ''
+    submissionID: PropTypes.string.isRequired
 };
 
 export default class HistoryTable extends React.Component {
     constructor(props) {
         super(props);
 
-        this.isUnmounted = true;
```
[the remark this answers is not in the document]
ebdabbs: It's unnecessary, since once the component instance is unmounted it will not be rendered again. `componentWillUnmount` should generally be used to do things like cancel API requests, stop timers, etc.

src/js/components/validateData/ValidateDataFileComponent.jsx
ebdabbs: Updated the `clickedReport` function to check for `this.state.canDownload`

PR #1260, opened 2020-01-27
Sprint 98: Staging -> Prod

alburde1:


dpb-bah:
* Refactored the overall site's router to allow upgrading the React versions.
* Removed unused node packages for a cleaner installation process.
* Upgraded React version to be compatible with modern versions of React.

PR #1439, opened 2020-10-05
Sprint-116 Release Notes (warmfix)

dpb-bah:


src/help/technicalHistory.md
```diff
@@ -1,3 +1,10 @@
+#### September 21, 2020{section=technical}
+
+In this release, here is a list of technical changes that may require infrastructure or database updates, or represents additional functionality.
+
+* Updated submission\_metadata’s “Last Modified” values to match what is displayed on the Submission Table.
+* Updated front-end behavior various DABS pages when a submission is reverting to prevent potential issues.
+
 #### September 4, 2020{section=technical}
```
alburde1, line 8: Can we start removing the `technical` tag for the old sections? Having the same section tag for multiple things doesn't help and we cleaned that up in a different PR so let's keep that going.

alburde1 (changes requested):

src/help/technicalHistory.md
dpb-bah: Yeah I was planning to fix that with the conflicts with the downmerging with subheader fixes. I can start it here.

alburde1 (approved):

PR #1695, opened 2022-07-25
Package Updates

alburde1:
**High level description:**

Updating some packages based on dependabot suggestions and eslint.

**Technical details:**

Only simple packages were updated in this PR, further update work to be done later.

**Link to JIRA Ticket:**

N/A

**Mockup**
N/A

Author: 
- Linked to this PR in JIRA ticket
- Scheduled Demo including Design/Testing/Front-end OR Provided Instructions for Local Testing above and in JIRA
- [x] Verified mobile/tablet/desktop/monitor responsiveness
- [x] Verified that this PR does not create any *new* accessibility issues (via Axe Chrome extension)

Reviewer(s):
- Design review completed


.eslintrc
```diff
@@ -25,18 +26,18 @@
         "indent": [2, 4, {
             "SwitchCase": 1
         }],
+        // require 4 spaces in JSX as well
+        "react/jsx-indent": [0],
+        "react/jsx-indent-props": [0],
```
dpb-bah, line 31: Why are we turning this off? And if justified, remove/update the comment associated with it.

dpb-bah (changes requested):

.eslintrc
alburde1: It's getting turned off because it seems to act oddly suddenly and looks for an extra tab where there shouldn't be one. It's in direct conflict with the regular indent, which covers all files (tested locally by adding an extra space in a file and it showed up in the linter even with these turned off). The current status, including comments, directly mirrors the USAS linter setting file. Did we want to deviate from that?

dpb-bah (approved):

